PR #20, opened 2022-01-26
fix weird sprite renaming bug

cedric-h:


leomcelroy:
Bug remains, will correct on dev branch.

PR #37, opened 2022-01-31
Add bug reporting

YodaLightsabr:
## Functional changes
- Adds the `REPORT_BUG` dispatch event
- Links the `REPORT_BUG` dispatch event to the bug button
- Adds the option to not copy the URL upon saving (disabled by default, only used for the `REPORT_BUG` dispatch event)

## Design changes
- Makes the images in buttons display pixelated instead of blurry

### DIFF
--- a/dispatch.js
+++ b/dispatch.js
@@ -137,8 +137,8 @@ const ACTIONS = {
       return (string.length > length ? string.substring(0, length - ending.length) + ending : string);
     }
     report['IP Address'] = await fetch('https://ifconfig.me/ip').then(response => response.text());
-    report['Dispatch Event Log'] = state.dispatchLogs.slice(0, 50).map(entry => truncate(JSON.stringify(entry, null, 4), 1000, '...')).join('\n\n');
-    report['Error Log'] = state.logs.slice(0, 50).map(entry => truncate(entry.stack || JSON.stringify(entry, null, 4), 1000, '...')).join('\n\n');
+    report['Dispatch Event Log'] = state.dispatchLogs.slice(0, 50).map(entry => truncate(JSON.stringify(entry, null, 4), 1999, '...')).join('\n\n');
+    report['Error Log'] = state.logs.slice(0, 50).map(entry => truncate(entry.stack || JSON.stringify(entry, null, 4), 1999, '...')).join('\n\n');
     report['User Agent'] = await fetch('https://ifconfig.me/ua').then(response => response.text());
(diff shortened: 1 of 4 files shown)


editor.css
```diff
@@ -248,3 +248,7 @@ body {
 
 
 
+
+button > img {
+  image-rendering: pixelated; /* Make button icons crisp */
```
maxwofford, line 253: Ooooh! TIL

dispatch.js
```diff
@@ -120,6 +122,47 @@ const ACTIONS = {
...
+      examples: state.examples,
+      error: state.error,
+      mouseX: state.mouseX,
+      mouseY: state.mouseY,
+      engineVersion: state.engineVersion,
+      previousID: state.previousID,
+      selected_asset: state.selected_asset,
+      name: state.name,
+      lastSaved: state.lastSaved,
+    }, null, 4), 50000, '...');
```
maxwofford, line 155: The character limit is 100k per field. Is there a reason we're truncating at 50k?

dispatch.js
```diff
@@ -120,6 +122,47 @@ const ACTIONS = {
...
+    notification({
+      message: "Generating a bug report... (3/3)"
+    });
+    let querystring = '';
+    for (const key in report) {
+      querystring += `&prefill_${encodeURIComponent(key)}=${encodeURIComponent(report[key])}`;
+    }
+    const url = `https://airtable.com/shrpcDFA5f9wEOSIm?${querystring.substring(1)}`;
+    window.open(url, '_blank');
```
maxwofford, line 164:
```suggestion
    const url = new URL('https://airtable.com/shrpcDFA5f9wEOSIm')
    for (const key in report) {
      url.searchParams.append(`prefill_${encodeURIComponent(key)}`, encodeURIComponent(report[key]))
    }
    window.open(url, '_blank');
```

I think this'll do the same thing, but with the built-in URL class

maxwofford (approved):
This looks great! Thanks for adding it. Added a couple comments, but looks good to merge. I'll bring it into my branch when I wake up

dispatch.js
YodaLightsabr: Oops! I forgot to update that.

maxwofford:
Screw it, this looks great so I won't wait. Merging!

YodaLightsabr:
Oops, I think the querystring thing is causing unexpected behavior. Image

Issue #14, opened 2022-01-25
show errors on the line they occur instead of below


cedric-h:
zach suggested this during a meeting and it just makes sense

maxwofford:
<img width="1151" alt="Screen Shot 2022-01-25 at 16 12 22" src="https://user-images.githubusercontent.com/5891442/151060553-3237a87a-5253-4ad5-a9b3-a7d3d23c0774.png">

maxwofford:
I'm not totally sure if this should have tag:bug, or tag:enhancement. Moving the error indicator to the line itself is an enhancement, but the error says it's on dispatch.js:88, which is incorrect (to the user, it's on line 7 & not in dispatch.js). Does this ticket also include fixing up the error message to mention the place the user's code errored?

YodaLightsabr:
I've done a bit of research and I have a few findings to report.

Because of how JavaScript `new Function` works, it checks for syntax errors, and since the line where it checks for the syntax error isn't in the actual code, it just says it's at `new Function()` and provides no more detail. However, when it's a runtime error, since it was found within the code, it shows you exactly where it is.

| Syntax error: | Runtime error: |
| ------------- | -------------- |
| ![Syntax error](https://yodacode.xyz/F81c4b3fa0e.png) | ![Runtime error](https://yodacode.xyz/Fc830b250dd.png) |

We might be able to get around the syntax errors by putting it through some kind of validator and showing the error from there if it finds one. For the runtime error, we would just need to parse the error string for the line number and column number, and then subtract the line by 2 (because it's starting at 3 for some reason) and the column number by 6 because it starts at 7.

leomcelroy:
Implemented this when possible with new error messages.

PR #40, opened 2022-02-01
Improved error messages

YodaLightsabr:
This includes improved error messages by stripping down the browser-given error message to only two things:
1. the stack trace (inside of the editor, not dispatch.js or any other gamelab files)
2. the type of error or message given by the browser.

This also adds the esprima library to check for errors ahead of time because the browser does not give you a stack trace for syntax errors. (Only runtime errors.)
This has been tested on the latest version of Chrome, and works fine. Fixes are needed for Firefox and Safari.

See #14 

Firefox: 
image

Safari:
image

### DIFF
--- a/dispatch.js
+++ b/dispatch.js
@@ -10,6 +10,7 @@ import { createSequencer } from "./sequencer/sequencer.js";
 import { playTune, loopTune } from "./tunePlayers.js";
 import uiSounds from "./assets/ui-sounds.js";
 import notification from "./utils/notification.js";
+import validate from "./utils/validate.js";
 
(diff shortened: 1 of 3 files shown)


YodaLightsabr:
It should now be fixed for Chrome, Firefox, and Safari. It might not work anywhere else, though, because error messages are different in every browser. Safari barely gives you any information.

maxwofford:
Thanks so much for adding this! Looking great. In the future, please merge into the dev branch.

Can we close https://github.com/hackclub/game-lab/issues/29 now?

PR #57, opened 2022-02-08
fix collision issue

maggie-j-liu:
i think this fixes the issue where the `collides` function is called before the objects actually collide

also fixes hitbox...
before: 
image
after: 
image


maxwofford:
Was this the issue you mentioned in our meeting @cedric-h?

PR #70, opened 2022-03-14
Fix share links to be Open Graph embed-compatible

cjdenio:


leomcelroy:
Not relevant after migrating to Sprig. We did add a serverless function to fetch thumbnails.

Issue #60, opened 2022-02-08
Prompt for name on share if user hasn't changed project name


maxwofford:
Brought up by @mikeymcfish. That'll help prevent 100s of projects called "name-here"

maxwofford:
This was added in #87

PR #87, opened 2022-03-24
Prompt rename on save for projects without names

maxwofford:
Users are prompted to name their projects before saving if they haven't yet...

https://user-images.githubusercontent.com/5891442/159982874-708b8da0-c4e9-4702-a33e-8d554a93bbaa.mov

Only triggers if a name hasn't been set & doesn't trigger on automatic saves (ie. bug report generation)

### DIFF
--- a/dispatch.js
+++ b/dispatch.js
@@ -193,7 +193,7 @@ const ACTIONS = {
     });
     const report = {};
     report["Engine Version"] = state.engineVersion;
-    await dispatch("SAVE", { type: "link", copyUrl: false });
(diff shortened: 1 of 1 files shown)


maxwofford:
Fixes https://github.com/hackclub/gamelab/issues/60

maxwofford:
@cedric-h over slack you mentioned #90 was merged in favor of this, but it doesn't seem like that addresses #60, which is the intent of this PR– this is unrelated to getting users to save more often.

PR #90, opened 2022-03-29
Save more, more often

cedric-h:


dispatch.js
```diff
@@ -307,6 +318,10 @@ const ACTIONS = {
     state.mouseY = mouseY;
     dispatch("RENDER");
   },
+  GENERATE_NAME({}, state) {
+    state.name = "gamelab-" + randString(3);
```
maxwofford, line 322: Won't this be nondeterministic? If we do this, won't we lose some utility of the MD5 in the ID?

cedric-h: Right, but this is only dispatched when a new project is created. The hash is still hashtastic and determinalicious.

PR #93, opened 2022-03-31
Implement anykey pressed or held

maxwofford:
This is a draft while figuring out the right way to implement this...

_Press [any] key to continue_

Screen Shot 2022-03-31 at 10 58 08

This PR implements checking for any keys being pressed or held. It uses the existing `heldKey()` and `pressedKey()` methods, but now giving them no argument checks for any key. You can also pass in "AnyKey" as the argument for the same effect.

```js
if (e.heldKey()) {
  console.log('A KEY IS PRESSED!!!')
}
if (e.heldKey("AnyKey") {
	console.log("Also means a key is pressed"
}
```

Before merge:
- [ ] Update docs
- [ ] Update examples(?)

### DIFF
--- a/Engine.js
+++ b/Engine.js
@@ -681,12 +681,18 @@ class Engine {
     return new Text(str, x, y, ops, this.textContainer);
   }
 
-  heldKey(key) {
-    return this._heldKeys.has(key);
+  ANY_KEY = "AnyKey";
+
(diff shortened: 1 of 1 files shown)


leomcelroy:
I like passing in no argument. I added this in the HTML-Export PR.

maxwofford:
Closing in favor of #97

PR #98, opened 2022-04-04
Document bundling Game Lab with Deno

cjdenio:
It works!

I considered documenting the use of [terser](https://github.com/terser/terser) (the raw bundle is around 1.6mb unminified), but decided it wasn't worth mentioning.


leomcelroy:
Thank you but we migrated to new Sprig system.

PR #101, opened 2022-04-06
Make buttons scrollable on mobile

YodaLightsabr:
Even though Gamelab isn't designed to support mobile, this PR makes the buttons list scrollable, allowing mobile users to run games.

Previously, if you were on a small device, the run button would be *just* out of reach.

Old buttons:
screenshot

New buttons:
screenshot
You can scroll the new buttons to access all of them.


leomcelroy:
Nice!

PR #162, opened 2022-07-22
Added support to take spacebar as an input

rithulkamesh:
Title says it all!!


leomcelroy:
Inputs have to match inputs on hardware, the Sprig Console. Discussed on Slack.

PR #195, opened 2022-08-07
fix(getTile): wrong variable use

farreltobias:
Fix getTile using `state.dimensions.height` insted of `state.dimensions.width` for x verification


leomcelroy:
Thank you!

PR #198, opened 2022-08-14
Normal Maze 

JayPlay71:


leomcelroy:
I love the random map generation Jay! I added a title and your name to the game.

I also think the game could use a few more elements to round it out. Right now if you just repeatedly hit the direction you want to move in then eventually the map will let you pass. Maybe if you added a timer or something else which penalizes the player for doing that then it would have more balanced game play. Nicky Case shared some great advice on game design with me a few weeks ago which I think is nice to keep in mind. Nothing should be purely positive or purely negative, everything should involve tradeoffs. If a player action is strictly good or bad then the strategy becomes clear, just be greedy. But if there is a tradeoff, then you have to really strategize. 

Thanks for the PR! I'll merge it but encourage you (and others) to continue improving the game.

leomcelroy:
Here is the form to get on the Sprig waitlist Jay!

https://airtable.com/appN9cSnDgORAM2bp/tblYLyzi8RySfkONg/viwMrglQ6x9pBSE0O?blocks=hide

PR #226, opened 2022-08-22
Create Spanish version of the docs

J-cordz:


leomcelroy:
Thanks for doing this Javi! Sort of hard for me to verify its correctness but I'll merge it in.

You should try making a game! We're really looking for feedback on that experience and then you'll be able to verify whether or not the docs work. Definitely open to changing the language (in Spanish and English) if it can be improved.

PR #271, opened 2022-08-28
Nomis!

recursiveforte:
Small Simon-like game I made


leomcelroy:
Nice Cheru! Really like the Simon remake, naming it Nomis makes me think I should put the pattern in backwards though. I think I've played versions like that before. 

The main enhancement I would make is add sound effects to each color. Otherwise this will be a great contribution to the game library!

leomcelroy:
Actually I see you have music but as you note it lags. What could we do to make it behave as desired?

leomcelroy:
Here's the form to get on the Sprig waitlist:

https://airtable.com/appN9cSnDgORAM2bp/tblYLyzi8RySfkONg/viwMrglQ6x9pBSE0O?blocks=hide

PR #273, opened 2022-08-28
alien

anshimathur0325:


leomcelroy:
Nice Anshi! I can see you put some thought into adapting Space Invaders to Sprig. Space Invaders was one of my early programs actually. I think this game is almost there but could use a little more polish.

I had one issue where the next alien just wouldn't show up. Maybe it's being shown offscreen? Here is a screenshot:

Screen Shot 2022-08-29 at 2 27 33 PM

I would also add a background to make it a bit more interesting visually. 

The biggest thing is that I feel the game needs more excitement or tradeoffs in gameplay. Like have some obstacle or danger to the player. The simplest change would just be to decrease the timer by a lot. Overall make it so I want to get a high score. I was trying to contrast the gameplay here to flappy birds. I think what makes flappy birds extra engaging is you're actively trying not to lose while trying to get a high score. In aliens right now there is only a high score but no obvious losing.

anshimathur0325:
Hi, I added more to my game. Please let me know if there is anything I need to fix!

leomcelroy:
Nice work incorporating the feedback! There is still a hack where you just move up vertically to where the missile needs to be then laterally smash into the alien. It seems to be the easiest route through each level which is why you may want to nerf the technique but overall this feels improved to me! I think you've earned your Sprig. Congrats!

You can order it here: http://sprig-order.hackclub.dev/

PR #285, opened 2022-08-30
Serve logo locally

maxwofford:


leomcelroy:
This won't work when the site is deployed. We might need a static folder for vercel right?

PR #297, opened 2022-09-03
Fix a typo in the help menu.

Omay238:
### afterInput(callback)

Runs after ever input event has finished being handled. Useful for tasks like checking win states:

Add Y after ever


leomcelroy:
Thank you! Nice catch.

PR #299, opened 2022-09-04
Update Sokoban+

Omay238:


leomcelroy:
Great work! I love these extensions and this is exactly the sort of thing we want to see people do. You turned the Sokoban game into something which can entertain you for some time now. You've definitely earned a Sprig!

You can fill out this form to order one: http://sprig-order.hackclub.dev/

PR #301, opened 2022-09-07
Scooter Game

LucasHT22:
Stop at the houses to know where the correct place is


leomcelroy:
The levels don't seem to progress for me. It also appears that the objective is hardcoded. It may be good to add some element of randomness to the game. Considering this I think the game is still incomplete. But this is also your second game! I love the sprites and playful feel to this and your previous game. I can tell you're spending some time making them, which is great so I think you've earned your Sprig. Congrats! You can order it here:  http://sprig-order.hackclub.dev/

I encourage to keep improving your games also!

cedric-h:
I really love the long strip level, it took me entirely too long to figure out 😅 

I couldn't see the little ✅ !

PR #312, opened 2022-09-07
Maze game

itamardavidyan:


leomcelroy:
Thanks for the submission to the gallery Itamar! It's a nice simple maze game. I think the easiest way to do walls is just full solid blocks but that wastes more space than the technique you used, which you implemented in a clean way. Aesthetically it would be nice if the walls lined up. You may need to create some more tile types for that to work. I'm referring to correcting this phenomenon 

Screen Shot 2022-09-08 at 11 30 02 AM

Otherwise a few more levels (that get more challenging) would make the game feel more complete. Once again really appreciate the PR. 

Also incase you were interested in the Sprig. We're only giving those to teenagers or younger, but you can still get your game added to the gallery.

itamardavidyan:
> Also incase you were interested in the Sprig. We're only giving those to teenagers or younger, but you can still get your game added to the gallery.

@leomcelroy I created this PR for my son.

And we plan to create more levels and make the walls lined up.

PR #315, opened 2022-09-08
You are the imposter...

LookItsAndy:
This game is a simple maze where you must find your way to the other crewmate in order to win as an imposter. My experience in coding this game took longer than I thought but as someone who has not used javascript before, I learned a lot about it.


LookItsAndy:
making new game

PR #314, opened 2022-09-08
shoot green blobs

Otterly-Otter:
This is fun!

PR #317, opened 2022-09-08
Classic Tank Battle

ayushgoyal7796:
Classic Tank Battle


leomcelroy:
This seems to be a copy of shoot green blobs which was submitted here: https://github.com/hackclub/sprig/pull/314

PR #320, opened 2022-09-08
killpigs

filiafobico:
You are a storm god and you want some pigs to make your dinner...

  So you need to choose and kill the rigth pig into each of 4 levels
  Which level the number of pigs increase
  With "a" you move left
  With "d" you move rigth
  With "j" you throw the ray

  Enjoy and good dinner :)


leomcelroy:
Great art and fun little narrative which comes across well. I think it could feel a little more game like if there was a memory component, perhaps if every level you had to complete all the preceding levels.

Welcome to the gallery! If your a teenager you can get your Sprig here: https://sprig-order.hackclub.dev/

PR #322, opened 2022-09-08
Remove empty tags from metadata.json

Neesh774:
image
Empty string tags caused a weird visual glitch, removed those tags


leomcelroy:
Awesome thank you!

PR #324, opened 2022-09-09
Create PASS_ALL.js

Shivang0:
@title: PASS_ALL
@author: Shivang0

Instructions:
1) Cover All Tiles
2) You cannot take step back. So, be careful
3) Move with aim to cover all boxes to win


leomcelroy:
Thanks for the PR! I quite like the puzzle games and it's fun to have new maze levels. 

I noticed a couple things that need some attention. Levels two and three both reference "w" which isn't defined anywhere so they just don't work. Also perhaps it's me being bad a playing but I don't see how level one is actually possible. Any block which is only open on one side has to be your last move but there are three blocks that meet this criteria just in the starting position. How is this level supposed to be solved?

Screen Shot 2022-09-13 at 11 54 42 AM

Shivang0:
Hello, I Pushed the previous version which was having issues. Uploaded the new version. Kindly check.

LucasHT22 (commented):
Good game! :)
Now it is possible to solve level 1,
Could it be more creative with levels 2 and 3? So it seems that the game is inverted, starts hard ends easy...

Shivang0:
It should be solvable now i guess.

LucasHT22:
Now yes! it is possible to complete the game, it has no bugs and it is original :)

PR #325, opened 2022-09-09
Trash_IT & Paint_IT

StuntStorm:
Hey there guys!

Here is my 2 New game for Sprig: **TRASH IT** & **PAINT IT**
----------------------------------------
**1. TRASH IT**
--------------------
Objective: **Delete the File**

> Put the Folder into the Trash
> Avoid the 💀 FOR YOUR OWN SAFETY!
> Press L for MAGIK
> Press K for Music
> As you progress, it will be more confusing 😏

--Controls--
> W to bring the mouse Up
> S to bring the mouse Down
> D to bring the mouse Right
> A to bring the mouse Left
----------------------------------------
**2. PAINT IT**
--------------------
Objective: **Make amazing Art using Sprig Console!**

--Controls--
> W to bring the mouse Up
> S to bring the mouse Down
> D to bring the mouse Right
> A to bring the mouse Left
> I to bring the mouse Black Color
> J to bring the mouse Red Color
> K to bring the mouse Blue Color
> L to bring the mouse Yellow Color


StuntStorm:
Added 2 More games : 
1. Trash IT 
2. Paint IT

Please check them and add it to the gallery.
Thank you,
Rizwan

leomcelroy:
More great submissions! I'm digging the theming.

Paint IT makes me feel like you're own your way to making a virtual console inside our virtual console.

I appreciate all the levels you made for Trash IT. Once again it's got this nice relaxing vibe and the puzzles are fun too. I think the death by skull effect may have a bug. What is the proper behavior? I thought that when the mouse comes into contact with the skull you go back to level 0. The way it's implemented though every skull has to be covered with the mouse. Here's the relevant snippet

```
const skulls = tilesWith(skull).length;
const target = tilesWith(skull, player).length;
if (skulls === target)  {
    tr = level = 0;
    const ds = levels[tr];
}
```

But perhaps it should be:

```

const skulled = tilesWith(skull, player).length;
if (skulled > 0)  {
    tr = level = 0;
    const ds = levels[tr];
}
```

StuntStorm:
> More great submissions! I'm digging the theming.
> 
> Paint IT makes me feel like you're own your way to making a virtual console inside our virtual console.
> 
> I appreciate all the levels you made for Trash IT. Once again it's got this nice relaxing vibe and the puzzles are fun too. I think the death by skull effect may have a bug. What is the proper behavior? I thought that when the mouse comes into contact with the skull you go back to level 0. The way it's implemented though every skull has to be covered with the mouse. Here's the relevant snippet
> 
> ```
> const skulls = tilesWith(skull).length;
> const target = tilesWith(skull, player).length;
> if (skulls === target)  {
>     tr = level = 0;
>     const ds = levels[tr];
> }
> ```
> 
> But perhaps it should be:
> 
> ```
> 
> const skulled = tilesWith(skull, player).length;
> if (skulled > 0)  {
>     tr = level = 0;
>     const ds = levels[tr];
> }
> ```

Thank you for the feedback!

Actually, When I figured that when there are 2 skulls in the scene, it won't work. I changed that into a feature of the game that the user had to figure out. So it would be like they are way too scared to hit the skull even though they can just bypass it when there is >1 skull 😂😂. Hence, the levels after 8 have way too many skulls.

Also yep I implemented your correction and yea it Fixed it! now that it is fixed should I rework the level or should I stick to the old script and keep it as it is?

leomcelroy:
I see. I was one of the people who became scared to hit skulls until the levels where it was unavoidable. I actually quite like that idea but some of the later levels become quite easy once you realize. In any case it's a design choice so up to you. These games are well made, once again nice work.

PR #326, opened 2022-09-09
Reversi

veehj:
Reversi, which one or two players can play.

### DIFF
--- a/games/reversi.js
+++ b/games/reversi.js
@@ -1,0 +1,505 @@
+/*
+@title: reversi
+@author: veehj
+*/
+
+const game = {
+  board: [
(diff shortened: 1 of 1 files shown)


leomcelroy:
This is very well done, nice work! The main thing that jumps out to me is that it can be confusing where you are allowed to go if you don't know the game already. It might be nice to have some sort of indicator tile of where valid moves are. At a minimum you could just write the rules in the comment at the top of the tile. I'll merge you in but definitely think you should address this point.

Welcome to the gallery! If you're a teen you can get your Sprig here: https://sprig-order.hackclub.dev/

PR #356, opened 2022-09-10
Sprig-Piano

Scott170c:
Simple Piano, Use A and D to point the arrow at a key and W to play the note
Sprite Editor was laggy but otherwise it was fun making it


leomcelroy:
I like the idea of making instruments with Sprigs. I think there is a small bug on line 152 where you reference song but mean to reference x. If you fix that I'll merge this in. Thanks for the PR.

Next I challenge you to make an instrument which is easy to play on the Sprig 😄 .

Scott170c:
> I like the idea of making instruments with Sprigs. I think there is a small bug on line 152 where you reference song but mean to reference x. If you fix that I'll merge this in. Thanks for the PR.
> 
> Next I challenge you to make an instrument which is easy to play on the Sprig 😄 .

Bug fixed, thanks for point that out! I will try making another instrument now that is easier to play on sprig :)

leomcelroy:
Great! Congrats on your Sprig: https://sprig-order.hackclub.dev/

PR #359, opened 2022-09-10
Tetris!

Neesh774:
Added Tetris!
It's pretty advanced, since it dynamically generates sprites and the map so pieces are different sizes and so using each one as a sprite would mess up collisions and being able to put multiple in the same general area. It also has music :)


leomcelroy:
Great work adapting this! It seems to run great, sounds and looks good. You're right it's a bit more advanced but the code is also written nice and simply. Good work! You already earned your Sprig so I won't send the link again but it's a welcome addition to the gallery.

Which classic will you do next?

PR #363, opened 2022-09-11
Added Mandelbrot

HenryBass:


leomcelroy:
If you didn't fill out the form yet you can do so here: https://sprig-order.hackclub.dev/

PR #364, opened 2022-09-11
Add Seven Ate Nine (789)

blobbybilb:
Hello, I'm a high schooler & I just found out about this a few days ago. I'm excited to try it, so here's what I made!

> Why is 8 scared of 7?
> Because seven eight (ate) nine!

---

A fun, fast-paced, 2-player game.

In Seven Ate Nine, one player plays as the 7, and the other player plays as the 9.

The goal for the 7 is to eat the 9. However, if the 9 get past the black line, the 8 will start chasing the 7, getting faster and faster.

I had fun making and testing the game; thanks for creating Sprig!


leomcelroy:
I like how you took a pun and turned it into an original game! The countdown is a nice example that other people may want to reuse too. I notice some sort of error that's related to trying to change properties on missing sprites. Here is one such error:

```
Cannot read properties of undefined (reading 'y')
```

You have to open up the console to see it.

Overall it's a nice original addition to the gallery. Welcome 😄 and congrats on your Sprig if your a teenager.

https://sprig-order.hackclub.dev/

Issue #362, opened 2022-09-11
Storage Abstraction for Games


LinusSkucas:
It'd be nice to have an API that games could use to access storage, whether that be on a physical device, or local storage in a browser. Two potential use cases for an API like this:
1. A game can write a high score to a file so it can keep track of it.
2. A player could import new levels into a game through files.
Abstracting this would make it so that games don't need to add extra logic to determine if it should be saved by the browser or on an actual file system.

UserJHansen:
I think this is possible in the js implementation on the Sprig as it is provided by kaluma, maybe a web interface is all that is required?

UserJHansen:
#365 should enable this

leomcelroy:
I took note of this in [this doc](https://github.com/hackclub/sprig/blob/main/docs/BACKBURNER.md) and will close this issue as not planned for now.

PR #365, opened 2022-09-12
Update to enable storage API

UserJHansen:


UserJHansen:
Fixes #362

leomcelroy:
Love that you're thinking about feature extensions and enabling persistent data is an interesting one. You can just access localStorage in the script anyway. You're right that we will want an abstraction layer to make it cross-platform but before we add that to the API let's get a demo running of what we may want it to do.

PR #370, opened 2022-09-14
Cube Snek

JavedTadavi:
I don't have any idea about sring 
But I'm try to create game but I don't have any idea to write spring code game 
I open help center and i learn and complete game 😃
All credit goes to Snek game creator


leomcelroy:
It's great that you're jumping into learning these things and I really appreciate sharing your game. I think it's not differentiated enough from the snake you were building from yet so I'm going to close this but keep at it and you'll have an original game in no time. One twist that came to mind was what if every time you grew longer your head switched to your tail.

PR #375, opened 2022-09-15
Fix most games.

Omay238:
I couldn't find out how to fix 9 puzzle. Sorry.


leomcelroy:
Nice corrections! How'd you find them all? It took me a while playing through every game.

PR #376, opened 2022-09-15
Fruit Appearence Bug Fixed

souvikpal2000:
Before Fruit was appearing on Snake Body due to which Head of the Snake get detached from its body. Now the problem is FIXED. In future if I encounter with any problem, I will surely fix it.

### DIFF
--- a/games/snake_game.js
+++ b/games/snake_game.js
@@ -157,23 +157,30 @@ const showScore = () => {
 }
 showScore();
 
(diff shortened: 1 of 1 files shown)


leomcelroy:
Thanks for updating!

souvikpal2000:
> Thanks for updating!

Hey, can I change the name of my game?

PR #390, opened 2022-09-17
Add color_domination game

adrianoapj:
This game was created at Sprig Jam! :)

### DIFF
--- a/games/color_domination.js
+++ b/games/color_domination.js
@@ -1,0 +1,233 @@
+/*
+@title: color_domination
+@author: adrianoapj
+*/
+
+/*
+W, A, S, D - to move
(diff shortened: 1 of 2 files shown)


SamDev-7:
Thanks for the PR! I really like this game! Interesting how you have to think ahead of the opponent to win the game. 
One suggestion I have would be to remember where the latest selection was for each player, so you don't have to move around as much.

LucasHT22:
I agree with @SamDev-7 , good game :)

leomcelroy:
I like how simple the game is. It seems that going first would be a huge advantage. Nice work! Thanks for coming to the Sprig Jam.

PR #393, opened 2022-09-17
Falling Blocks

thanosan23:
Dodge the falling blocks for as long as you can. Use A and D to move.

I have experience coding in Python, but I do not have as much experience with JavaScript.

### DIFF
--- a/games/Falling Blocks.js
+++ b/games/Falling Blocks.js
@@ -107,16 +107,20 @@ const obstacleUpdate = setInterval(() => {
   }
   
   for(let i = 0; i < obstacles.length; ++i) {
-    if(obstacles[i][1] >= 11) {
+    if(obstacles[i][1] == 11) {
       toRemove.push(i);
     }
   }
   for(let i = 0; i < toRemove.length; ++i) {
     if(obstacles[i][0] != getFirst(player).x) {
-      clearTile(obstacles[toRemove[i]][0], obstacles[toRemove[i]][1]);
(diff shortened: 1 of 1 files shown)


SamDev-7:
Thanks for the PR, interesting game! One minor suggestion I have is that the block don't seem to actually touch the ground, they disappear before reaching the ground. Try tweaking your code, so they reach the ground and then disappear, let us know if you need any help!

thanosan23:
I fixed the game so the obstacles reach the ground before disappearing.

SamDev-7:
That works! There are still some minor issues I just noticed, if you quickly move after being hit by a block, the block stays there. Another very rare issue, if you move into a block just before it gets removed, it removes the player as well. I have an idea how those two could have occurred, let us know if you need any help. Happy hacking!

thanosan23:
Is there a way to remove all tiles in the last row without removing the player?

SamDev-7:
You should be able to check the type with
```javascript
if (sprite.type != "p")
```

thanosan23:
Thank you! I have fixed the bugs.

LucasHT22:
Great game!
I would merge this if I had the permissions :)

LucasHT22 (commented):
Game over sound effect reminded me of Chrome Dino haha

leomcelroy:
It's an elegant game. Welcome to the gallery! If you're a teen congrats on your Sprig. You can order it here: https://sprig-order.hackclub.dev/

PR #397, opened 2022-09-19
Fix metadata.json for reactor.js

SamDev-7:
Fixes duplicate entry for [`reactor.js`](https://github.com/hackclub/sprig/blob/main/games/reactor.js) by @sampoder.
Changes filename to `reactor` so it does not 404.


sampoder:
Thank you!

PR #401, opened 2022-09-22
Infinite Maze

pertark:
Randomly generated maze game! Use WASD or IJKL to move.


SamDev-7 (commented):
Thanks for the pr! Random generation is so cool!

One thing, can you rename the title/file, so it is only one word? Something like `infinite_maze`.
After that, it can be merged. :D

pertark:
It should be all good now @SamDev-7

SamDev-7:
Awesome! 

If you're a teen. You can order your Sprig here: https://sprig-order.hackclub.dev/ for free! :D

PR #415, opened 2022-09-29
alien_invaders

stati30241:
Simple clone of the classic space invaders game.

Use the A and D to move the player around.
Use the W or I to shoot.
Press J to replay.


LucasHT22:
The game is simple but very good!

PR #418, opened 2022-09-29
Delete test.html

sampoder:
If we're not using this, might be worthwhile removing to keep the repo clean


leomcelroy:
Reasonable! Thanks.

PR #419, opened 2022-09-30
Simon

RileyCz:
This is a 2x2 version of the game "Simon". Basically, tiles will light up in a specific order which you then have to recreate. The goal is to copy the pattern correctly for as long as possible. When you mess up, your score will appear briefly and a new game will start. Highest score possible is 99, higher than that and it will not be able to display your score properly.


LucasHT22:
Very good game! It's fun!

PR #421, opened 2022-10-02
Tower of Hanoi

nicholasz2510:
### Simple Tower of Hanoi game I threw together in about half an hour :)

Questions: 
* Is there any way to modify the Z coordinate of a sprite?
* Can you have more than 4 sprites on the same tile?


SamDev-7:
Thanks for the PR! Cool game! The controls were a bit finicky, but I got the hang of it!

Maybe it would be nice to show which ring is "selected" and will be moved. A step counter showing how many moves you have taken would also be really cool.
As a challenge, you could also try making a second level with 4 poles and 4 rings!

Feel free to ask here if you get stuck!

LucasHT22:
> ### Simple Tower of Hanoi game I threw together in about half an hour :)
> Questions:
> 
> * Is there any way to modify the Z coordinate of a sprite?
> * Can you have more than 4 sprites on the same tile?

@nicholasz2510  about your questions:
* The sprig is 2d so we usually use only x and y, of course there are games that give the impression of 3d, but there are cases and cases
* Not on the map, maybe if they're not solid and can move it's possible

nicholasz2510:
I revamped the controls to avoid locking the player into a move in case they misclick, and also added a move counter.

Also @SamDev-7 I wasn't able to find a good way of having 4 rings because that would mean 5 sprites on a same tile. I did however make it so you start off with just 2 rings and then move on to 3

SamDev-7:
Awesome! 
If you're a teen congratulations on your Sprig. You can order it here: https://sprig-order.hackclub.dev/

nicholasz2510:
Awesome, thanks for the merge!!

leomcelroy:
You can modify the z-order by changing the order you define sprites in setLegend.

The constraint on number of sprites per a tile comes from the sprig console itself. We could increase it if it's too low.

PR #422, opened 2022-10-02
Fixed: Repeatedly pressing shoot key after dying throws error

amaan-ahmad:
This fix is for the game: `Shoot green blobs`

On repeatedly pressing "J" key for shooting even after dying throws the following error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'playback') on line 439 in column 9. 
Open the browser console for more information.
```

Fixed it in the commit below 👍


LucasHT22:
Ok!

PR #428, opened 2022-10-04
floppyMatch

Gaurav-Ban22:
In floppyMatch, you control a pointer that can move from left to right. A floppy disk appears at the top of the screen, and you have to match it. You match it by moving the pointer with a and d and selecting a floppy disk top and bottom with i (that match the disk on the top). After you select two parts, the program checks whether the selections are correct. If they are, you get a point, but if they aren't, then you get a strike. After ten points, you win, but after 3 strikes, you lose and your progress is reset. (if you accidentally select a part that you don't want to, press k and your chosen parts should reset.) You can also press j to restart or play again. There is also a timer (it serves no use other than for fun).

I have been coding for the last 4-5 years, but mainly backend and other applications in languages like java, python, and cpp, as well as c# for unity game development.  I haven't really delved into web development or any other subsection needing javascript or typescript other than one time I tried a web framework, so I didn't have that much experience in js before starting this project. This project did help me learn a little bit more about js syntax, and the game-making library is pretty cool! I have only ever made games in Unity before and the gamemaking framework/library was really fun to work with; I had fun with this project


LucasHT22 (commented):
I didn't understand your game very well... could you explain it to me? What do the numbers on the screen mean?

Gaurav-Ban22:
Oh, so I forgot to write this in the pr description
The first number is how many matches you got right (score)
The second is how many strikes you have 
The third is how much time it took. I just committed something to update the timer every second.

When the second number hits 3, the timer resets and the score resets.
When you hit 10 points, the game ends and you win.
I apologize if my explanation wasn't sufficient.

LucasHT22:
> Oh, so I forgot to write this in the pr description The first number is how many matches you got right (score) The second is how many strikes you have The third is how much time it took. I just committed something to update the timer every second.
> 
> When the second number hits 3, the timer resets and the score resets. When you hit 10 points, the game ends and you win. I apologize if my explanation wasn't sufficient.

no need apologize!

besides the Omay question, is the sound "okay" really empty?

Gaurav-Ban22:
@Omay238 Yes, you have to choose the bottom (an oversight on my part ill try to debug) and yes, not knowing what you have selected is part of the challenge.

@LucasHT22 okay is an unused variable I was planning on using but never did. For optimization and cleanup purposes I should probably delete that. Thank you for reminding me!

I also made the timer more interactive and update every second in a new commit.

LucasHT22:
nice! just don't forget to add your changes ;)

Gaurav-Ban22:
Okay!

LucasHT22 (commented):
good changes! Could you add the instructions at the top of the code? After that I merge your game :)

Gaurav-Ban22:
I added instructions. Thank you!

LucasHT22:
Congrats!

PR #432, opened 2022-10-06
fix small "getting started" typo, d -> a

bellesea:
up and left should be "w" and "a" instead of "w" and "d"


LucasHT22 (commented):
Good eye! really, to the left is A and not W

PR #433, opened 2022-10-07
Add files via upload

tigerflame777:
Maze: This is a small maze game with 8 different levels increasing difficulty. I think this game would be a great addition to the sprig as it is simple and fun. I have only just started to code using JavaScript but I have coded a bit using python, c++, and scratch.

### DIFF
--- a/Maze.js
+++ b/Maze.js
@@ -1,0 +1,227 @@
+/*
+@title: Maze
+@author: Akeell 
+*/
+
+const player = "p";
+const wall = "w"; 
+const teleport = "t"; 
+const block = "b"
+
+setLegend(
+  [ player, bitmap`
+................
+................
+.......000......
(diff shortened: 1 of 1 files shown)


Omay238:
It's a pretty fun game, but some of the levels seem kind of just thrown together, some levels just have holes that you can't reach, but if that is the style you're going for, that's cool. Besides that, I think that this game could go on the gallery.

Omay238:
I feel like the main level with that problem is level 4, and also, I think you could make your own art for the player, but that's a lower priority.

Omay238:
Alright, this looks good. Do you want it merged now, or do you want to do anything else with it? Either way, if you're a teen congratulations, you got a Sprig. You can order it here: https://sprig-order.hackclub.dev/

PR #438, opened 2022-10-08
Candy Rush!

Tasbia:
Try to get the candies in the mouths.
-No pushing candies with other candies; there are ways to position your character to not have to do this to win -The black blocks are walls, which your character cannot pass through.

### DIFF
--- a/games/Candy Rush.js
+++ b/games/Candy Rush.js
@@ -1,0 +1,170 @@
+/*
+@title: Candy Rush
+@author: Tasbia Uddin
+*/
+
+
+const player = "p";
+const candy = "b";
+const mouth = "g";
+const wall = "w";
+
+setLegend(
+  [ player, bitmap`
+................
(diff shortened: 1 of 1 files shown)


Omay238:
It's fun, the art is good, but there is only one level. I feel like adding more would be really important if you want to get a sprig.

Omay238:
Nice! I'll merge this now.

LucasHT22:
I love the pun on the name of the game haha

Tasbia:
Ah whoops, sorry! I had my starting level not set at 0.

LucasHT22:
Sorry, but I don't understand what you mean, could you explain better?

Omay238:
> Sorry, but I don't understand what you mean, could you explain better?

I had previously put a comment on, saying there was one level, but they had the level variable set to some number that wasn't 0, likely for testing purposes.

PR #445, opened 2022-10-12
Hectic Hockey

FrozenThunder25:
Its a very straight forward game similar to frogger where you must reach the end of the map without colliding with the enemy. I have quite little experience with coding with only fairly basic knowledge with Python


leomcelroy:
Your art is really wonderful!

PR #446, opened 2022-10-12
puzzle_runner

ShanSu8:
In my game, you have to navigate through a maze while pushing a box until you reach the goal point.


LucasHT22 (commented):
Thanks for PR! Good game! I like game remix :)

PR #451, opened 2022-10-13
Add error about Web Serial API

recursiveforte:
Add an explicit error on pushing to a device if the browser does not support the Web Serial API


leomcelroy:
Thank you!

PR #477, opened 2022-10-24
minceraft

tildezero:
it may or may not be based on a popular game, w to jump, s/d to move side to side, your object is to collect all the emeralds to open the nether portal, and the lava kills you

loosely based on a similar [game](https://github.com/79t/thegame) i made for my CS class

I feel like I should add more to the game, but I'm not really sure what to add


LucasHT22:
Thanks for PR!

LucasHT22:
Was your objective for the spriteto float?
image

LucasHT22:
I think you could change the main sprite, be creative :)

LucasHT22:
Could you add your references in the code header?

cedric-h:
If you use `setBackground`, you can get rid of the blank white spaces in your map :) Check the docs for details!

I couldn't figure out the second level -- the emeralds didn't seem to work?

Could you make the gravity more ... gravity-y? Might be worth looking at [this game](https://editor.sprig.hackclub.com/?file=https://raw.githubusercontent.com/hackclub/sprig/main/games/mini_maze_and_puzzle.js), I think it had really great gravity!

> I feel like I should add more to the game, but I'm not really sure what to add
ooh, what if you took a note from [falling block](https://editor.sprig.hackclub.com/?file=https://raw.githubusercontent.com/hackclub/sprig/main/games/Falling_Blocks.js) and added things to dodge on some of the levels? :eyes:

So far, great work, can't wait to see what you come up with :)

tildezero:
thank you guys so much for the suggestions, im working on implementing them rn, but i do have some questions
> I couldn't figure out the second level -- the emeralds didn't seem to work?

the emeralds are kind of weird because i have them in afterInput, so i think you have to actually move onto one instead of fall to get it (and also it was deleting the first instance instead of the one you were on until now)

> If you use setBackground, you can get rid of the blank white spaces in your map :) Check the docs for details!

I tried to do this, but it sets the background for all maps, is there a way to not do that?


> ooh, what if you took a page from [falling block](https://editor.sprig.hackclub.com/?file=https://raw.githubusercontent.com/hackclub/sprig/main/games/Falling_Blocks.js) and added things to dodge on some of the levels? 👀

i actually had blazes that moved in the old game, but i havent implemented them yet

SamDev-7:
@tildezero How's the game going? 

> the emeralds are kind of weird because i have them in afterInput, so i think you have to actually move onto one instead of fall to get it (and also it was deleting the first instance instead of the one you were on until now)

Try putting all the emerald stuff in a function and calling that in the gravity, it might work.

> I tried to do this, but it sets the background for all maps, is there a way to not do that?

You can change the background with `setBackground(spriteKey)`. You can use if statements to detect which level it is to change the background.

Let us know if you need anything!

leomcelroy:
It seems you need to make more changes before you're ready to get your Sprig. I'm going to close this PR for now. Feel free to reopen if you want to continue working on your game.

PR #489, opened 2022-11-02
Update README.md

bellesea:


Omay238:
not leo, but the link that says "hack club" leads to the editor. is that intentional?

bellesea:
> not leo, but the link that says "hack club" leads to the editor. is that intentional?

oop no it's not, good catch! thank you :)

PR #506, opened 2022-11-07
inconsequential leveldungeon

LtPeriwinkle:
Collect keys to unlock the portal and escape the randomly-generated dungeon. Don't fall into a trap. The cake is not a lie.

WASD to move, J to start game (or to reset if you end up in an impossible dungeon, which could theoretically happen).

If it's too hard I can nerf it.
The death sound and item-get sound are mario and zelda, respectively, but I wrote them by hand (and they aren't exactly identical) so it should be fine.

I'm a pretty experienced programmer, but this is the most JavaScript I have written probably ever (and the most spritework I've ever done as well).


LucasHT22:
Thanks for PR!

LucasHT22 (commented):
Great game!

LucasHT22 (approved):
Congrats!

LtPeriwinkle:
Thanks!

PR #509, opened 2022-11-07
Rename bertblaster.js to Bert_Blaster.js

ungarmichael:
The game cannot be found in the gallery hence why I renamed it.


LucasHT22:
ok, I made a mistake when adding it to metadata.json, I could just change it there, but with this space in the name it's much better. thanks:)

PR #514, opened 2022-11-08
Add Conways Game of Life

the456gamer:
add thumbnail
add game

> The Game of Life, also known simply as Life, is a cellular automaton devised by the British mathematician John Horton Conway in 1970. It is a zero-player game
https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life

includes debug views if the view is a certain size (so text aligns correctly)
as well as two speeds, a single step mode, and a way to clear everything

there is some helper js in the bottom that ensures the timers etc are cleared up when the user restarts the code

this is still in progress


the456gamer:
Ok, should be good to review

the456gamer:
Ok, should be good to review

LucasHT22 (approved):
great game! Congrats!
you can get your sprig here! https://sprig-order.hackclub.dev/

(ref: https://hackclub.slack.com/archives/C02UN35M7LG/p1667916819490189)

PR #537, opened 2022-11-12
fix bug where it would kill the player twice

duck123acb:
I have already submitted this game, just want to update it


LucasHT22 (approved):
Ok!

PR #541, opened 2022-11-12
Update virus_plowing.js

duck123acb:
Update virus_plowing.js to include the ability to restart, fixing placement of text, and fixing a bug where the player’s choordinates would go off the screen, but the image wouldn’t.


LucasHT22:
ok!

PR #559, opened 2022-11-14
update virus_plowing.js

duck123acb:
When I spawn a new virus, it will now get destroyed if it spawns on another sprite; before my if statements for destroying the new virus were wrong.

I also added some comments


SamDev-7:
Awesome! Great work.

PR #584, opened 2022-11-26
Fix the audio on lovelock

UserJHansen:
This is just a PR that fixes the bug in Love Lock allowing you to press K more than once and the audio overlapping. It also deletes the unused horrorbg track.


LucasHT22:
Looks good to me, but first @StuntStorm  must agree

leomcelroy:
It's okay to merge PRs that fix bugs. Thanks for helping @UserJHansen

StuntStorm:
Hey all,

sorry for the delay..was busy with my research paper. Thanks for fixing the
bug ✨✨

PR #588, opened 2022-11-29
Simple rendering

leomcelroy:
Reverting back to canvas rendering. Addresses some minor rendering issues such as limit of 4 sprites per tile. Also makes sprig engine more easily usable as a library.


leomcelroy:
I would love help checking every game before we merge this in. @Omay238 @LucasHT22 @Neesh774 @SamDev-7

LucasHT22:
I'll check now

LucasHT22:
all games are playable, in yarn test only `Snake_4_2 ` returns error in properties `x ` and `y`, I've tried everything but I couldn't solve it

Omay238:
@LucasHT22 Can you put a screenshot in the github thread? I have been unable to access Slack recently. I can try to debug it.

LucasHT22:
ok! 
image

Omay238:
I think I fixed it. I changed every time it looks for `something.x` to check that it exists. I also changed == -> === and != -> !== to just be correct JS syntax.

LucasHT22:
Very good @Omay238 ! Thanks

leomcelroy:
It's because there are random inputs so the test is not exhaustive or repeatable. But the test also looks at the map not the render. So really we have to manually check the games.

PR #637, opened 2022-12-16
Add strong typings

khrj:
This PR adds a `global.d.ts` file that can be used when developing a Sprig game using TypeScript.
I've covered all of the methods available in the documentation, but perhaps there are some I'm not aware of.
Even though keys, bitmaps, maps, colors and tunes are still strings, I'm using branded types to differenciate them

- [x] Level Design
    - [x] `setLegend`
    - [x] `setBackground `
    - [x] `setMap`
    - [x] `setSolids`
    - [x] `setPushables`
    - [x] `width`
    - [x] `height`
- [x] User Input
    - [x] `onInput`
    - [x] `afterInput`
- [x] Sprites and Tiles
    - [x] `getTile`
    - [x] `tilesWith`
    - [x] `addSprite`
    - [x] `getAll`
    - [x] `getFirst`
    - [x] `Sprite.remove` 
- [x] Text
    - [x] `addText`
    - [x] `clearText` 
- [x] Music
    - [x] `playTune`
    - [x] `Playback.end` 
- [x] Debugging 
    - [x] `getState`  
- [x] Tagged Template Literals
    - [x] `bitmap`
    - [x] `map`
    - [x] `color`
    - [x] `tune`

### DIFF
--- a/global.d.ts
+++ b/global.d.ts
@@ -8,11 +8,10 @@ declare global {
 	type Tune = string & { __brand: "tune" }
 
 	type Sprite = {
-		_type: string
-		_x: number
-		_y: number
-		dx: number
-		dy: number
+		type: string
+		x: number
+		y: number
+		remove: () => Sprite
 	}
 
 	type InputKey = "w" | "a" | "s" | "d" | "i" | "j" | "k" | "l"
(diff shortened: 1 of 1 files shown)


leomcelroy:
Thanks for doing this! Some people have actually requested it.

LucasHT22:
@khrj @leomcelroy Sorry, but how would saving files work if they are all saved in .js by default?

khrj:
> @khrj @leomcelroy Sorry, but how would saving files work if they are all saved in .js by default?

This isn't meant to work with the web editor, you can write code locally and then compile it to js to run it on the editor. So the .d.ts file doesn't interface with the editor at all

LucasHT22:
ok! thanks.

PR #644, opened 2022-12-19
Add run, hide and watch url params

khrj:
- `run=true` can be set to run the code given in file immediately 
- `hide=true` can be set to not show the editor, just the run window
- `watch=X` can be set to refresh the file param every X milliseconds

Each of these params require `file=` to be set

Example:
```
https://editor.sprig.hackclub.com/?file=http://localhost:3000/build.js&run=true&hide=true&watch=1000
```

This will reload the file from http://localhost:3000/build.js every second (if it has changed), while also immediately rerunning it, and not displaying the editor

### DIFF
--- a/dispatches/init.js
+++ b/dispatches/init.js
@@ -62,7 +62,10 @@ export async function init(args, state) {
     // These params only make sense when running from a file
 
     if (getParam("run")) dispatch("RUN")
-    if (getParam("hide")) document.querySelector(".code-container").remove()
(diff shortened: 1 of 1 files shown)


leomcelroy:
Neat idea! I was a little reluctant to add run by default because the executed js isn't sandboxed very well.

leomcelroy:
But the whole app is clientside so it's not a huge worry.

Issue #496, opened 2022-11-05
security issue: cross site scripting in file load


the456gamer:
Demo: open url and click run

https://editor.sprig.hackclub.com/?file=https://gist.githubusercontent.com/the456gamer/c539813d9902185578d799f3aaed9772/raw/d152b2b85ccdaf4d3067a906c50c04e0c5ae76e8/my-cool-sprig-game.js

file source: https://gist.githubusercontent.com/the456gamer/c539813d9902185578d799f3aaed9772/raw/d152b2b85ccdaf4d3067a906c50c04e0c5ae76e8/my-cool-sprig-game.js

this abuses the fact that the JavaScript isn't sanitised before being run (another demo i had was using a fetch() to send/receive information)
and that arbitrary scripts can be loaded with the ?file= parameter

alhardwarehyde:
@grymmy can you reproduce this?

grymmy:
Will attempt to reproduce thus tomorrow.

kognise:
This isn't relevant anymore.

PR #665, opened 2022-12-22
fix security bug (#496)

Omay238:
only auto run file when trusted and warn user if not trusted


Omay238:
#496

Omay238:
not yet functioning

LucasHT22:
image
Great work

PR #667, opened 2022-12-23
fix: Delete duplicate maze_runner.js.

shanecelis:
`maze_runner.js` is a duplicate of `Maze_Runner.js`. It causes warnings on macOS when checking out the repo.

```
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'games/Maze_Runner.js'
  'games/maze_runner.js'
```


LucasHT22:
Ok!

PR #682, opened 2022-12-28
Add specifications for screw length in steps

duck123acb:
While I was assembling my sprig console, I was a little bit confused on which screw length to use in step 2.
So I specified which length to use in both contexts (steps 2 and 6)


exu3 (approved):
That's helpful, thanks! 

For more clarity, it might be worth mentioning that the longer screws come in the same bag as the spacers.

PR #688, opened 2022-12-29
changed tune-text converter to output ints (truncated)

the456gamer:
(for durations)
this matches the tune parser on the web version, so no infomation is lost however, in hackclub/spade, the durations seem to be read as floats, so some infomation maybe lost as each int represents 1/500th of a second, the loss seems acceptable

additionally, the varibles names were swapped to match the format used in parsing, now that it matters (as the tone is different from the duration, even if they were both previously inserted verbatim)

## About your game

no game created / updated

> Thanks for PR!


grymmy:
@the456gamer is this still code that you feel should land?  It is difficult for me to understand how this could be tested - how did you approach doing so?

the456gamer:
With the new sprig migration, I'm not sure on what was changed / migrated
I don't have time to work on this ATM, so I'll mark as draft, and check again soon

From what I remember, there were several changes, including fixing the variable name, since they were flipped in one case , but it didn't matter since they were the same type

I don't think these changes should be noticeable? But it significantly increases readability if the text tune

grymmy:
> With the new sprig migration, I'm not sure on what was changed / migrated
> I don't have time to work on this ATM, so I'll mark as draft, and check again soon
Kk thx for the update, understood.  When you do get the chance we will move this quickly towards merging.

grymmy:
I am closing this due to inactivity.  If you feel strongly that this has been closed in error, please reopen with a comment.

PR #718, opened 2023-01-11
Hot-Diggity-Dog

TheScientist101:
# Your checklist for this pull request

## About your game

What is your game about? 
A hotdog is trying to escape the invasion of pineapple, but needs to stay hydrated with mustard. Can you help him survive?

How to play your game?
WS to move up and down. J to restart at any time.
## Code
Check off the items that are true.
- [x] The game was made using the [Sprig editor](https://editor.sprig.hackclub.com/).
- [x] The game is placed in the in the [`/games` directory](https://github.com/hackclub/sprig/tree/main/games).
- [x] The code is significantly different from all other games in the [Sprig gallery](https://sprig.hackclub.com/gallery) (except for games labeled "demo").
- [x] The game runs without errors. 
- [x] The name of the file/game contains only alphanumeric characters, `-`s, or `_`s.
- [x] The game name is not the same as the others from [gallery](https://sprig.hackclub.com/gallery)

## Image (If an image is used)

- [ ] The image is in the [`/games/img` directory](https://github.com/hackclub/sprig/tree/main/games/img).
- [ ] The name of the image matches the name of your file. Example: `sokoban_plus.js` -> `sokoban_plus.png`.

## Experience
I started coding in middle school with CS50. Since then I have coded a chat app, competed in the NASA App Development Challenge, competed in CS UIL and now I am trying game development with Sprig.

PR #719, opened 2023-01-11
Fix the locked in a box bug for the hot-diggity-dog game

TheScientist101:
This was my game in pr #718. I realized that I had a bug later. I hope this is the right way to make a pr for this.


LucasHT22 (approved):
Looks good to me!

PR #722, opened 2023-01-12
Fix the score overflow bug for real this time in the game hot-diggity-dog

TheScientist101:
The game hot-diggity-dog had a bug where when the score reaches 100 or more, the number would show partially on the other side of the screen. This fixes that.


LucasHT22 (approved):
Ok!

PR #749, opened 2023-01-23
feat: Added Slow Motion for cube_escape.js

GalaxyGamingBoy:

## Update
Updated `cube_escape.js` and added a `How to play` section, a changelog and most importantly slow motionCan be enabled / disabled by setting the `slowMotion` variable to the corresponding value. i.e. Set it to true if you want slow motion or to false to disable slow motion. Slow Motion works by looping *after* a player movement.


LucasHT22 (approved):
LGTM!

PR #772, opened 2023-02-03
Fixed a minor bug (`Find the Suspect`)

unordentlich:
* Fixed bug that still allowed selecting defendants even though the game was already over


LucasHT22 (approved):
LGTM!

PR #780, opened 2023-02-06
Falling Water bugfix, and a new thumbnail.

Captainexpo-1:
I fixed a bug where the water would get stuck on top of the walls and I added a new thumbnail.


LucasHT22 (approved):
Ok!

PR #788, opened 2023-02-12
Bannai Snake

ahoysal:
# About

Solve puzzles while following calligraphy rules! Move yourself to the star, while making sure to have a valid board when you reach it. A valid board is defined by:

1. No paths touching only by a diagonal (no 2x2 squares with diagonals being the same color). Two lines touching through a right angle is okay.
2. No "thick" paths that are 2+ tiles wide (no 2x2 squares that are all black or all white)

That's it! It's a little tricky at first, but once you start you'll pick it up in no time. You can double back on where you've already gone, unlike the game snake, so you can move onto tiles adjacent to you, no matter if they are black or white.
One last rule: gold bordered tiles can't be moved through, so you'll need to move around them.
There are four levels, see if you can beat them all!

Inspired by the rules of [Bannai Caligraphy](https://en.wikipedia.org/wiki/Kufic#Square_Kufic) (also called Square Kufic)

## Controls

WASD to move up, left, down, and right.
J to reset the puzzle or advance when you've finished.

# Checklists

## Code
Check off the items that are true.
- [x] The game was made using the [Sprig editor](https://editor.sprig.hackclub.com/).
- [x] The game is placed in the in the [`/games` directory](https://github.com/hackclub/sprig/tree/main/games).
- [x] The code is significantly different from all other games in the [Sprig gallery](https://sprig.hackclub.com/gallery) (except for games labeled "demo").
- [x] The game runs without errors. 
- [x] The name of the file/game contains only alphanumeric characters, `-`s, or `_`s.
- [x] The game name is not the same as the others from [gallery](https://sprig.hackclub.com/gallery)

## Image

- [x] The image is in the [`/games/img` directory](https://github.com/hackclub/sprig/tree/main/games/img).
- [x] The name of the image matches the name of your file. 

> Thanks for PR!


LucasHT22 (changes requested):
interesting game! just out of curiosity, could you explain me more about this calligraphy?

ahoysal:
Yeah sure! It's a type of Arabic calligraphy derived from Kufic, which is a really old style with an angular look. Square Kufic was first used in 12th century Iran on buildings because it's easier to do that type of text with bricks (_banna'i_ comes from the term for masonry in Farsi) and is still featured on their flag to this day. Because of its blocky appearance, it sacrifices some legibility in exchange for a structured appearance. It follows the same rules as the game, but sometimes bends them a little, for example having the text be in non-rectangular canvases, using rotated lines, and having different colors. (Source: [Wikipedia](https://en.wikipedia.org/wiki/Kufic#Square_Kufic))

Let me know if you have any more questions!

LucasHT22 (approved):
Interesting! Looks good to me! Congrats:)

Issue #815, opened 2023-02-21
Newline characters are not rendered properly on Sprig


WojtekWidomski:
Text after `\n` is moved to left. This bug does not exist in web editor.

Example code:
[https://editor.sprig.hackclub.com/?id=ae01a926735fc1b67e5e1833bf93f9bc](https://editor.sprig.hackclub.com/?id=ae01a926735fc1b67e5e1833bf93f9bc)

**Editor:**
![screenshot](https://user-images.githubusercontent.com/98206524/220355822-1e8d34b2-1aef-42f1-a07e-d5107e1b7269.png)

**Sprig:**
![newline_bug](https://user-images.githubusercontent.com/98206524/220356857-04f6e0cf-b97b-4a7e-89be-875f28699e1f.jpg)

PR #854, opened 2023-02-24
Update randomworld

WojtekWidomski:
- Game is now faster and uses less RAM
    - World is stored as array of Uint8Arrays.
        - only 4 bits of int are used to store block type
        - 5th bit is 1 when there is an enemy. This allows to check if there is an enemy without iterating over all enemies.
    - Enemies are now stored in 8x8 chunks. This allows to update only enemies close to player.
    - `generateWorld` now modifies `world` global variable instead of returning new array. This uses less memory when restarting game.
- Add option to disable sound
- Add loading screen
- Update sounds
- Workaround #815
- Make thumbnail square


LucasHT22 (approved):
LGTM!

PR #793, opened 2023-02-14
the_hungry_mouse

KaiEtkin:
# Your checklist for this pull request

## About your game

My game is about a mouse trying to get all the cheese, while avoiding the moustraps.

How to play your game?

Use WASD to move

## Code
Check off the items that are true. 
- [x ] The game was made using the [Sprig editor](https://editor.sprig.hackclub.com/).
- [x ] The game is placed in the in the [`/games` directory](https://github.com/hackclub/sprig/tree/main/games).
- [x ] The code is significantly different from all other games in the [Sprig gallery](https://sprig.hackclub.com/gallery) (except for games labeled "demo").
- [x ] The game runs without errors. 
- [x ] The name of the file/game contains only alphanumeric characters, `-`s, or `_`s.
- [ x] The game name is not the same as the others from [gallery](https://sprig.hackclub.com/gallery)

## Image (If an image is used)

- [ x] The image is in the [`/games/img` directory](https://github.com/hackclub/sprig/tree/main/games/img).
- [ x] The name of the image matches the name of your file. 

> Thanks for PR!

PR #882, opened 2023-03-01
(re)add #793

the456gamer:


the456gamer:
Closing in favour of #881
Which I didn't see. Oops

Issue #964, opened 2023-03-14
Undo in editor erases everything after page reload


SamDev-7:
Pressing undo (control/command z) after a reload results in a blank file, causing you to lose all your progress.

## Expected behavior
To go back to the previous edit and if no previous edit is saved, to do nothing.

## How to reproduce
1. Open the editor.
2. Type something.
3. Wait for it to save.
4. Repeat 2-3 a couple of times.
5. Press undo, it works.
6. Reload the page.
7. Press undo.
8. Your entire code is erased. 

As there is no file history, if you do not have a backup, you risk losing your game forever if you do not redo in time.

grymmy:
This is 100% reproducible on my end.  Thank you for the instructions.

grymmy:
It appears that when loading up the initial code in the editor using CodeMirror, we dispatch a new (undoable) modification of the text.  Thus, when the user hits 'undo', the editor undoes that text addition, which results in all code being removed.

Relevant code:
`if (props.initialCode) editor.dispatch({ changes: { from: 0, insert: props.initialCode } })`

Looking into prepopulating the editor now without causing a new transaction to be put into the history stack.

SamDev-7:
Seems to be fixed with the recent changes.

PR #968, opened 2023-03-14
Do not initialize the text in the editor using an undoable transaction

grymmy:
Fix to #964


sampoder (approved):
neat!

grymmy:
> neat!

Thanks!

PR #1030, opened 2023-05-14
Throw console error if glyph doesn't exist.

Omay238:
Closes #441

PR #1033, opened 2023-05-16
Let space be rendered

ahoysal:
PR #1030 does not allow the space character to be rendered, throwing an error instead. This PR allows space as a character to be rendered.


Hrushal-Nikhare:
Facing the same issue

SamDev-7:
@Omay238

Omay238:
Good catch. I originally recognized it, but totally forgot.

PR #1066, opened 2023-06-10
Patch for "Two Birds, One Stone"

t3mp0r4ry23:
This PR features a patch/bugfix for *Two Birds, One Stone*, which fixes an issue in which the score would not be reset after failure.


LucasHT22 (approved):
LGTM!

PR #1070, opened 2023-06-14
Fix broken sprite at fungi-frog-maze.js

jzaleta:
The game file was using an old sprite that causes issues when loading on the gallery 😅


LucasHT22 (approved):
good catch!

PR #1111, opened 2023-08-12
Sprikipedia, The Sprig wiki and encyclopedia - Under development

LucasHT22:
Hello Sprig World, this is Sprikipedia!

> What is Sprikipedia?

Sprikipedia, is the Sprig wiki and encyclopedia.

* So the wiki on GitHub is over?

> Nope, the /sprikipedia is a begginer friendly and more beautiful version of Wiki on GitHub, Sprikipedia also supports more languages. However, things like Runbooks, which don't interest beginners/general audiences or are more team focoused, will remain on the GitHub wiki.

* How people can contribute to Sprikipedia?

> When it's released, anyone can edit/add/fix our docs simply with a PR here in the repo.

---
Sprikipedia is currently under development, we have many decisions, meetings and conversations ahead.

---

As Graham (@grymmy) asked in one of our meetings, here is the PR Draft.


src/global.css
```diff
@@ -88,4 +88,44 @@ a:hover {
 
 .cm-lineNumbers > .cm-gutterElement[err-line] {
 	background: #ecb2b2!important;
-}
\ No newline at end of file
+}
```
grymmy, line 91: It might be safer to move all sprikipedia css into its own file to ensure changes you make do not side-effect other preexisting pages...

src/global.css
```diff
@@ -88,4 +88,44 @@ a:hover {
...
+
+/* Bad code by Lucas */
```
grymmy, line 93: remove comment if in jest, or at least specify why it is bad...

grymmy: ohhh I see - this was just to section off your changes (definitely think it makes more sense to put in separate file).

.gitpod.yml
```diff
@@ -0,0 +1,11 @@
+# This configuration file was automatically generated by Gitpod.
```
grymmy, line 1: Is it important for this file to be checked in?

src/layouts/pt-br/categorias.astro
```diff
@@ -0,0 +1,23 @@
+---
```
grymmy, line 1: Correct me if I'm wrong, but the plan would be to remove this and just have one language-agnostic layout, right?

src/layouts/pt-br/docsbr.astro
```diff
@@ -0,0 +1,22 @@
+---
```
grymmy, line 1: This also would be removed due to it not being language-agnostic in layout definition, right?

grymmy (changes requested):

src/global.css
LucasHT22: Ok! Will do

.gitpod.yml
LucasHT22: Nope, this is just a file to my source control operates, but need to delete before the merge

src/layouts/pt-br/docsbr.astro
LucasHT22: I thought I removed these files, I'll remove

V205Arduino:
Cool! is it sprigipedia or Sprilipedia?
Will it be available at the Hackclub website?

LucasHT22:
> Cool! is it sprigipedia or Sprilipedia? 

Right now it's Sprikipedia

> Will it be available at the Hackclub website?

Will be avaliable at the Sprig website:)

.gitpod.yml
grymmy: for future reference, if you don't want this checked in and don't want to deal w/ it in the commit status, you should add it to the `.gitignore` file!

public/translations/EN.json
```diff
@@ -0,0 +1,50 @@
...
+    "home.docs.about.description": "The history, purpose, everthing",
+    "home.docs.brand.name": "Brand",
+    "home.docs.brand.description": "All branding material for download",
+    "home.dark.info": "Come to the dark side!",
+    "about.head.title": "About Sprikipedia - Sprikipedia",
+    "about.content.title": "About Sprikipedia",
+    "about.blockquote.quote": "Imagine a world in which every single person on the planet is given free access to the sum of all Sprig knowledge. That's what we're doing.",
+    "about.blockquote.citation": "Lucas, rephrasing Jimmy Wales quote from Wikipedia's about page",
+    "about.content.p1": "Sprikipedia is a encyclopedia.",
+    "about.content.p2": "Sprikipedia's purpose is to benefit readers by containing information on all branches of Sprig, Written collaboratively by members of the <a href="https://hackclub.com">Hack Club Community</a>. Since its creation, as a <a href="https://github.com/hackclub/sprig/wiki">wiki page</a> in <a href="https://github.com/hackclub/sprig/">our GitHub repository</a>, on March 13, 2023 by Graham, it has grown to cover all things Sprig.",
```
grymmy, line 25: It seems like you would be mixing the presentation/UI layer with the localization layer here, which is probably not desirable.  Suggest keeping these cleanly separate.

LucasHT22: done!

grymmy (approved):
Lucas,

I believe this PR is ready to land in its current state, provided the following is correct (which I believe it is - double-check me on this):

1. The existing sprig website does NOT link to this page
2. This PR does not side-effect any other page in the sprig editor

Provided the above is true (and you can test #2 out a bit just by navigating around and testing), please land the code such that you can do smaller bite-sized PRs (more easily reviewable changes) on top of it

G

PR #1131, opened 2023-09-03
Sprikipedia, The Sprig wiki and encyclopedia - Under development (3)

LucasHT22:
PREV. : #1111 & #1129


jakjakob:
are you also planning on adding what error codes mean and maybe also common issues and how to fix them?

LucasHT22:
> are you also planning on adding what error codes mean and maybe also common issues and how to fix them?

Thanks for the suggestions! I'll

grymmy (approved):

PR #1147, opened 2023-09-22
Add stop button

### LINKED ISSUE #1146: Stop running button
would be useful to not having it running all the time while coding

Tiger-The-Cat:
Adds a stop button that appears when a sprig game is run in the web editor
image

closes #1146.


grymmy:
@leomcelroy are you okay with this change?  I do not believe it was vetted product-wise

leomcelroy:
As I mentioned in the #1146 

I do approve this feature. But by replacing the play button it requires you to hit the button twice to rerun your game. Does that also mean that shift+enter now toggles between playing and pausing the game loop. 

I would instead suggest adding the pause button next to mute.

grymmy:
@Tiger-The-Cat Can you please chime in here when this PR is ready for review and in compliance with @leomcelroy 's request for how this feature should function?  Feel free to assign this PR back to me when this is done...

Tiger-The-Cat:
Thank you! I didn't know shift+enter was a method to run the program, so I didn't implement a shift+enter toggle. I assume shift+enter shouldn't stop the game loop, as that would conflict with the flow of being able to restart the game in one action?

I've pushed the other change for now!🥳

P.S. space was a little tight, so I had to shrink the "Actual Sprig screen is 1/8" / 160×128 px" to just "Sprig screen is 1/8" / 160×128 px"

Tiger-The-Cat:
@grymmy I can't seem to assign anyone to this PR
image

grymmy:
> @grymmy I can't seem to assign anyone to this PR
> image

Assigned to Leo as I believe that was the request? Next time I'm at my laptop I'll add u with triage perms

grymmy:
@leomcelroy I guess this is rdy for review

jakjakob:
even though it is already changed, wouldn't it make more sense that the stop button is at the same spot as the run button (similar to a play/pause button in music player), instead of near the mute button (that technically works like a switch; so it would be a switching button near a single-action button (bad user experience/counter-intuitive (??)). The only downside of putting the stop button at the same spot as the run button that I can think of is that you need to click one time more if you leave the game running while editing the code (correct me if I am wrong).

Tiger-The-Cat:
Thanks for the comment, @jakjakob! I think @leomcelroy is concerned that the extra step you mentioned will interfere with the current workflow for developing and running sprig games. I think I agree with him, being forced to double click every time to run code quickly becomes pretty aggravating.

grymmy:
Plz resolve merge conflicts and then ping me for review when desired

Tiger-The-Cat:
Merge conflicts should be resolved now👍

JosiasAurel:
@Tiger-The-Cat please can your resolve the merge conflicts

PR #1150, opened 2023-09-25
add metrics on api functions

JosiasAurel:
- get metrics from sprig api functions using Astro middleware


grymmy (approved):
lgtm as long as you've tested this and it appears to work properly

PR #1155, opened 2023-09-29
add timing metrics to sprig

JosiasAurel:


grymmy (approved):
My only concern is fears that some path usages might result in cardinality but I am sure you will take that into consideration, same stuff we've had to deal w/ on other services...

PR #1160, opened 2023-10-05
Sprikipedia, The Sprig wiki and encyclopedia - Translation to Portuguese (WIP)

LucasHT22:
PREV. : #1111, #1129, #1131

PR #1163, opened 2023-10-07
Sprikipedia, The Sprig wiki and encyclopedia - Making more beautiful (WIP)

LucasHT22:
PREV. : #1111, #1129, #1131, #1160


grymmy (approved):
Suggest you make smaller more focused PRs in the future that narrowly address singular issues.  My understanding is that this still does not link Sprikipedia to the main page, and thus I am fine w/ this landing (there are some issues still, obviously)

PR #1166, opened 2023-10-10
fixing level 10 spider position

usermaatresetepenre:
hey i placed the spider in a position that is not the starting point for my debugging. now i fixed spider position 🖤


LucasHT22 (approved):
LGTM!

PR #1185, opened 2023-10-19
add dark mode to the editor

JosiasAurel:
This changes:
- adds dark mode to the sprig editor
- that's it!


grymmy:
@JosiasAurel can you confirm this pr changes no behavior by default, and any change in this setting would be opt-in?

grymmy (approved):
lgtm as long as this has 0 effect by default

PR #1219, opened 2023-11-06
Update codemirror.tsx to fix dark mode not saving upon reload

### LINKED ISSUE #1204: Tutorial in editor doesn't follow theme (dark mode)
**Describe the bug**
On the tutorial panel in the Sprig editor, if you have dark theme enabled the background will still be white and in light mode. The help tab on the same panel does switch to dark mode.

**To Reproduce**
1. Go to the Sprig gallery and enter a tutorial
2. Switch to dark mode
3. See that the tutorial panel stays on light theme

**Expected behavior**
I expected the tutorial panel to switch to dark theme.

**Actual behavior**  
The tutorial panel stays on light theme.
**Screenshots**
image

### LINKED ISSUE #1214: Code Editor doesn't follow theme on refresh
**Describe the bug**
Upon refresh or when I open the code editor from the My Games page, the editor page changes to dark theme but the code editor doesn't. I can only get the code editor to switch to dark theme when I click the "Dark" button at the top of the editor page.

**To Reproduce**
1. Go to 'My Games' page
2. Click on a game
3. Switch to dark theme
4. Refresh
5. Notice code editor is light theme

**Expected behavior**
I expected the code editor to always be dark theme when I have dark theme set.

**Actual behavior**  
The code editor switches to light theme on refresh. 
**Screenshots**
image

BrightTheBackpack:
closes #1214 
closes #1204


grymmy:
@JosiasAurel please review.

JosiasAurel (approved):

src/components/codemirror.tsx
```diff
@@ -43,6 +43,7 @@ export default function CodeMirror(props: CodeMirrorProps) {
 			effects: StateEffect.reconfigure.of(restoreInitialConfig())
 		});
 	};
+	setEditorTheme();
```
kognise, line 46: This is running every render cycle, it should be inside a useEffect... which it is, but that apparently isn't working? I haven't troubleshooted myself but maybe a useSignalEffect is more appropriate? Or something else is happening that's making CodeMirror not change the theme early.

kognise (changes requested):

src/components/codemirror.tsx
BrightTheBackpack: useEffect isn't ran when the page first loads, which is causing the editor to be in light mode when reloading. It's only ran on updates. We can try useSignalEffect.

BrightTheBackpack: I added one outside of the useEffect, which will run when the page first loads.

BrightTheBackpack: The rest of the screen is changed to dark mode because it is stated directly in the html code, using the imported isDark varible

BrightTheBackpack: @kognise

grymmy:
@kognise please respond to the OPs changes

src/components/codemirror.tsx
JosiasAurel: @BrightTheBackpack you can remove the hanging `setEditorTheme` and instead add `editorRef` in the useEffect that sets the editor theme

JosiasAurel: @BrightTheBackpack you can add a remove the hanging `setEditorTheme` and instead add `editorRef` in the useEffect where the editor theme is set.

JosiasAurel: alternatively you can put the extensions in a list and append the dark theme if `isDark.value` is true before returning that.

BrightTheBackpack: I will try your suggestions, but I'm just wondering what's wrong with the current implementation?

BrightTheBackpack: I put editorRef and removed the extra setEditorTheme

kognise (approved):

PR #1242, opened 2023-11-15
move from /games to /games/img

LucasHT22:


LucasHT22:
just to squash the commits

PR #1251, opened 2023-11-19
Add note about browser and Linux to UPLOAD.md

polypixeldev:
This PR adds a note to the uploading guide about using a supported browser and the `dialout` group if on Linux.


grymmy:
@leomcelroy please give your opinion here re: desired browser support - is it true that product-wise we only require support for "Edge or Chrome"?

leomcelroy:
We rely on web-serial which is only supported in chromium browsers and I don't see a workaround other than asking for a the installation of a native bridge for browser to serial port communication.

PR #1254, opened 2023-11-21
add timing to sprig db operations

JosiasAurel:
time database operations


grymmy (changes requested):
Please change the signature of timedOperation to take in a metric key and perform instrumentation internally to the helper method - this will simplify / reduce LOC

grymmy (approved):

Issue #1253, opened 2023-11-20
Autosave does not save as expected when losing network connectivity and then regaining it


grymmy:
**Describe the bug**
The editor does not gracefully handle saving data when resuming network connectivity.

**To Reproduce**
1. Open an existing game from Your Games in the editor.
2. Add one space, forcing an auto-save - observe save be successful
3. Disconnect your network connection (turn off wifi, for instance)
4. Add one space.  Observe a failure to save.
5. Resume your network connection (turn on wifi).

**Expected behavior**
The new space should be successfully saved shortly after resuming network connectivity.

**Actual behavior**  
"Error saving" popup repeatedly flashes and never saves successfully (even after a full minute or two from what I've observed - until you make another change).

**Screenshots**

https://github.com/hackclub/sprig/assets/43505167/84d29503-fd75-4b9f-add6-979d81d1a91b

JosiasAurel:
fixed since #1255

PR #1255, opened 2023-11-22
fix: autosave to retry every two seconds after a failed attempt

JosiasAurel:
This change addresses #1253.
It will retry saving every two seconds after a failed save attempt so that users do not lose their work.


src/components/big-interactive-pages/editor.tsx
```diff
@@ -39,34 +39,49 @@ const heightMargin = 130
...
+				persistenceState.value = {
+					...persistenceState.value,
+					cloudSaveState: 'ERROR'
+				} as any;
+			}
+		}
+
+		await attemptSaveGame();
+		while (isError) {
+			await attemptSaveGame();
```
grymmy, line 74: this and subsequent line probably should be swapped

src/components/big-interactive-pages/editor.tsx
```diff
@@ -39,34 +39,49 @@ const heightMargin = 130
...
+
+				persistenceState.value = {
+					...persistenceState.value,
+					cloudSaveState: 'ERROR'
+				} as any;
+			}
+		}
+
+		await attemptSaveGame();
+		while (isError) {
```
grymmy, line 73: The usage of this 'isError' variable is slightly evil.  Couldn't the error state just return from the attemptSaveGame method instead of managing mutable state, and wouldn't that be cleaner?

i.e. while(!await attemptSaveGame()) {

grymmy (changes requested):
Beyond my specific comments please clean up the rest of the PR - it does seem to have some unrelated + formatting changes.

src/components/big-interactive-pages/editor.tsx
```diff
@@ -57,29 +56,29 @@ export const saveGame = debounce(800, (persistenceState: Signal<PersistenceState
...
 				} as any;
+				return false;
 			}
 		}
 
-		await attemptSaveGame();
-		while (isError) {
-			await attemptSaveGame();
+		let error = !await attemptSaveGame();
+		while (error) {
```
grymmy, line 72: can't we just do: while(!await attemptSaveGame()) ?

grymmy: also - what is the consequence of this loop running for a long time?  This will not block the user's UI, will it?  Can this end up resulting in multiple error loops happening in parallel?  Sort of concerned that we might want to make this more asynchronous (a separate timer mechanism)...?

JosiasAurel: aha i pushed some more updates already

JosiasAurel: this has not blocked any more functionality in the editor as far as I've tested

src/components/big-interactive-pages/editor.tsx
```diff
@@ -47,26 +47,39 @@ let lastSavePromise = Promise.resolve()
...
+				persistenceState.value = {
+					...persistenceState.value,
+					cloudSaveState: 'ERROR'
+				} as any;
+				return false;
+			}
+		}
+
+		while (!await attemptSaveGame()) {
+			if (await attemptSaveGame()) break;
```
grymmy, line 72: this line is unnecessary

grymmy (approved):

Issue #490, opened 2022-11-03
Throw an error if users redefine a library function


cedric-h:
Tomb of Terrors has its own function called `getGrid`.

cedric-h:
note: raycast is guilty of this as well
<img width="175" alt="image" src="https://user-images.githubusercontent.com/25539554/200034234-00cb7d04-ea02-4f70-91c6-df463afadc72.png">

alhardwarehyde:
@kognise is this issue still relevant? If so can you clarify what the issue is? Thanks!

kognise:
The device firmware and the browser engine are inconsistent— on the device if a library function is redefined the code won't run, but this isn't the case on browser because the game is just evaluated from a function body and library functions (specified as parameters to the game function) are simply shadowed.

An alternate possible solution to this is to mirror the execution of games in the browser on device; perhaps placing the game code in a function closure and passing lib functions as args. This could be done from Spade or from the browser uploader, not sure which would be better but possibly Spade. Care must be taken to make sure this doesn't screw up error messages or make them more confusing. Perhaps we should be plopping the function body as a `JSON.stringify`'d string into `new Function`?

JosiasAurel:
resolved since #1266

PR #1266, opened 2023-11-28
throw error if user redefines library function

JosiasAurel:
addresses #490 

Will throw an error if a user redefines a library function

image

### DIFF
--- a/src/lib/engine/index.ts
+++ b/src/lib/engine/index.ts
@@ -58,7 +58,19 @@ export function runGame(code: string, canvas: HTMLCanvasElement, onPageError: (e
 	code = `"use strict";\n${code}`
 	
 	try {
-		parseScript(code)
+		const program = parseScript(code, { loc: true })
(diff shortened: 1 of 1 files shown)


src/lib/engine/index.ts
```diff
@@ -58,7 +58,19 @@ export function runGame(code: string, canvas: HTMLCanvasElement, onPageError: (e
 	code = `"use strict";\n${code}`
 	
 	try {
-		parseScript(code)
+		const program = parseScript(code, { loc: true })
+		for (let item of program.body) {
+			if (item.type === "FunctionDeclaration" && Object.keys(api).includes(item.id!.name)) {
```
grymmy, line 63: This might be auto-optimized, but would suggest keeping the API keys precached just once on startup instead of potentially recreating them repeatedly in this loop.

grymmy (changes requested):

Issue #1151, opened 2023-09-25
Sometimes orpheus isn't on the homepage.


leomcelroy:
Sometimes the page loads and looks like this.

![Screenshot 2023-09-25 at 1 24 48 PM](https://github.com/hackclub/sprig/assets/27078897/234079d3-c201-456d-85ae-267cfb9472b9)

When it should have orpheus in the corner.

![Screenshot 2023-09-22 at 5 35 37 PM](https://github.com/hackclub/sprig/assets/27078897/1ee67314-2457-484a-bef2-7af29acd5e74)

This ^ image shows the hovered state.

LucasHT22:
I didn't even know it existed haha

BrightTheBackpack:
I couldn't really reproduce this issue. Does it disappear at random?

BrightTheBackpack:
Looks like it might have something to do with the dimensions of your screen. If the width of the screen is too small, it will not appear. I will investigate further and try to find a fix.

grymmy:
@BrightTheBackpack yes, I can confirm - I see it appearing and disappearing based on the aspect ratio of the screen - in portrait, it doesn't show up (probably offscreen?)

grymmy:
@LucasHT22 any chance you could look into this one?  if not, please assign it back to me and I'll find a better home

LucasHT22:
Looks like this is on purpose:
![image](https://github.com/hackclub/sprig/assets/88567622/9f45287b-b987-4e16-845b-f2d0bd6ef4be)

JosiasAurel:
@leomcelroy do we still want this to always be shown on the homepage? 
Why is this not visible on some displays @kognise ?

JosiasAurel:
@leomcelroy do we still want this to be shown on the homepage?

grymmy:
@JosiasAurel @LucasHT22 given @leomcelroy opened the bug somewhat recently I think we should just fix the issue and make it display bottom-left as expected

PR #1280, opened 2023-12-06
always show orpheus on the homepage

JosiasAurel:
addresses #1151


grymmy (approved):
I do wonder if this was originally done due to prevent some other sort of other artifacting - but lets let Orpheus shine ig

PR #1292, opened 2023-12-12
add sort by release date and sort by name in ascending/descending

### LINKED ISSUE #1289: add method to change order of sorting
It would be nice to have a way to switch sorting in the gallery from chronological to alphabetical. I have been relying on alphabetical sorting until now for playing through all the games, but even outside of that it would be nice to have more than one way to sort them.

JosiasAurel:
add chronological/alphabetical sort order to sprig gallery

moved the gallery component to React for more reactivity

resolves #1289

### DIFF
--- a/src/components/design-system/input.tsx
+++ b/src/components/design-system/input.tsx
@@ -9,6 +9,8 @@ interface InputProps {
 	class?: string | undefined
 	maxLength?: number
 	bind?: Signal<string>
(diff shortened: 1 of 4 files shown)


grymmy (approved):
Seems pretty reasonable.  Did also suggest in slack some tweaks to the ux - use your discretion re: that

PR #1294, opened 2023-12-14
author sprig runbook

JosiasAurel:


grymmy (approved):
a step in the right direction

LucasHT22:
I know that I shouldn't intrude in this, but Wouldn't it be better to keep this in /wiki?

JosiasAurel:
@LucasHT22 you're not intruding at all. We should probably have it there as well tbh I didn't know that existed :D

LucasHT22:
:) some time ago Graham said that would be better runbooks be in wiki bc it's faster to update

grymmy:
As long as easily accessible via github + added as slacker resource I'm 😊

grymmy:
Can link from wiki add Readme and slacker - all to same source of truth

PR #1316, opened 2023-12-19
improve search

JosiasAurel:
- include author in search parameter
- filter showed items using block/none display


grymmy (approved):
Seems simple enough and straight-forward

PR #1323, opened 2023-12-21
Multiple gallery fixes

grymmy:
1. Load game content (code + images) from local (server-side) instead of github - WYSIWYG
2. Fixes metadata issues resulting in 404's (all entries in metadata.json containing len > 0 caused 404's on client.
3. Fixes client-side thumbnail caching (every character pressed in search bar would redownload ALL images every time)

Overall this makes the gallery load quicker and be snappier, far less network bandwidth utilized


JosiasAurel (approved):
lgtm

Issue #1194, opened 2023-10-23
Confusing error + page freezing "Level must be rect"


zachlatta:
**Describe the bug**

**_Two bugs_**

_First bug:_

When helping someone on their game, we got the error `Level must be rect.`. After much searching, we discovered that a character was accidentally deleted from a `bitmap` declaration.

This error message is confusing because it was a `bitmap` declaration that was wrong, not a `map` declaration.

_Second bug:_

Opening the bitmap that is missing a character in the pixel editor freezes the webpage.

**To Reproduce**

1. Go to https://sprig.hackclub.com/share/CdUtqTkENBsCgrwABkWk
2. Try running the game
3. See the error `Level must be rect.`. Note that this is a confusing error, and that the line numbers aren't helpful / are wrong. The actual error is on line 347 of the code.
4. Go to line 347 and click `bitmap` to open the sprite editor. Note that your browser is now frozen and the tab crashes.

**Screenshots**

<img width="529" alt="Screenshot 2023-10-23 at 6 10 46 PM" src="https://github.com/hackclub/sprig/assets/992248/a2b7866e-ca50-40d3-ba95-592a20d4b0a7">

grymmy:
It may be this is related to #1091 ?

Issue #1040, opened 2023-05-20
Include line and column number for invalid character errors


SamDev-7:
Users can choose to add text to their game with `addText`. However, the font has a limited support for special characters. An error will appear when the user tries to character add text with unsupported characters. This change was implemented in (#1030) and can be seen in the [Sprig Engine](https://github.com/hackclub/sprig-engine/blob/e5e3c0c179ddee62b78d74753d836ab550b84b6e/src/base/text.ts#L17).

The example of the current error looks as follows.
```
Error: Character $ is not in the font. It will be rendered incorrectly.
```

As a quality of life change, it would help to include information about where the invalid character is.

An example of how this could look:
```
Error: Character $ is not in the font. It will be rendered incorrectly.
    at <game>:39:16
```

grymmy:
@SamDev-7 I am sorry for the slow response on this issue - however, I do not see sufficient information here for this issue to be actionable.  Can you please update the description and title of this issue to be more clear re: exactly what the feature request is - because I read this and while I could jump to conclusions, I'd rather not...

Please assign this back to me once you've added info here to make this actionable...

SamDev-7:
I have updated the description, let me know if further clarification is needed.

grymmy:
@leomcelroy please weigh in on your approval/disapproval re: this feature request.  If you approve, do please specify the full set of characters you feel we should support.  If you do not approve, plz close w/ commentary.

leomcelroy:
These characters are supported

 !\"#%&\'()*+,./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ\\^_-`abcdefghijklmnopqrstuvwxyz|~¦§¨©¬®¯°±´¶·¸ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ÙÚÛÜÝÞßàáâãäåæçèéêëìíîïñòóôõö÷ùúûüýþÿĀāĂăĄąĆćĊċČčĎĐđĒēĖėĘęĚěĞğĠġĦħĪīĮįİıŃńŇňŌōŒœŞşŨũŪūŮůŲųŴŵŶŷŸǍǎǏǐǑǒǓǔˆˇ˘˙˚˛˜˝ẀẁẂẃẄẅỲỳ†‡•…‰⁄™∂∅∏∑−√∞∫≈≠≤≥◊
 
 I do support improving the error messages to include error location when possible.

PR #1378, opened 2024-01-10
update sprig package version

JosiasAurel:
the sprig engine recently had changes addressing #1194 & #1040 

with this package update, these issues should no longer be a problem


grymmy (approved):
I approve of this, but per our slack chats @JosiasAurel we should carefully orchestrate the release of this between the web editor and spade firmware, etc...

PR #1384, opened 2024-01-14
Fix wording of hardware development kit paragraph

polypixeldev:
Was just looking at the Sprig site and noticed a couple grammar issues in this paragraph on the home page.


leomcelroy:
Looks good to me, thank you!

PR #1409, opened 2024-01-24
Remove No Game Found Text ||  Fixed multiple Plagiarism Report

DevIos01:
Removes the No Plagiarism Found text and only posts if it detects

Check if the comment has already been sent so it doesn't repost when workflow runs again


JosiasAurel (approved):
lgtm

Issue #1091, opened 2023-07-17
Run Sprig games in a web worker


recursiveforte:
Currently, Sprig games run in the main thread, meaning any freezes occurring in the Sprig game (like an infinite while loop) will freeze up the editor as well; this could be fixed by moving all game logic to a web worker that's terminated & recreated on each run.

I came across this issue this week when working on a game & having my editor freeze & delete my work after every unsuccessful run.

recursiveforte:
I can work on this if I have bandwidth later this week.

grymmy:
@recursiveforte do you still wish to look into this?  Seems like a reasonable feature

grymmy:
I firmly believe this should be considered a bug and not a feature request, and thus am reclassifying it as such.  The editor shouldn't break when someone accidently writes an infinite loop.

leomcelroy:
I reiterate that I strongly recommend implementing infinite loop checking with ast analysis as described by Replit. I've implemented the web-worker alternative and it brings a large number of challenges and alternative issues.

grymmy:
> I reiterate that I strongly recommend implementing infinite loop checking with ast analysis as described by Replit. I've implemented the web-worker alternative and it brings a large number of challenges and alternative issues. 

This will be taken into consideration

JosiasAurel:
The editor will now report potential infinite loops to prevent freezing since #1413

PR #1413, opened 2024-01-29
add infinite loop detection

JosiasAurel:
the sprig editor will throw an error if
1. there are more than 2000 iterations in a loop AND
2. the loop takes > 1.5s to complete

image

addresses #1091 

add environment variables to Vercel deployment
- [x] `PUBLIC_MAX_ITERATIONS`
- [x] `PUBLIC_MAX_LOOP_TIME_MS`

### DIFF
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -8,6 +8,23 @@ and sharing them with others. You can watch [this video](https://www.youtube.com
 
 Run games by hitting the `Run` button or pressing `Shift+Enter`.
 
+### Infinite loop detection
+The Sprig editor will automatically insert a heuristic in your `for`, `while` and or `do-while` loops to detect potential infinite loops.
+Code looking like
+```js
+while (condition) {
+  // do stuff
+}
+```
+will become
+```js
+startTime = performance.now()
+while (condition) {if (++_loopIt > 2000 && performance.now() - startTime > 1500) throw new RangeError("Potential infinite loop")
+  // do stuff
+}
+```
+Note that all original line numbers in your code will be preserved.
+
 ## Getting Help
 
 If this is your first time using Sprig, try playing through the [tutorial](https://sprig.hackclub.com/gallery/getting_started). From there, we suggest hacking on any of the [current games](https://sprig.hackclub.com/gallery) or starting from scratch.
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -8,23 +8,6 @@ and sharing them with others. You can watch [this video](https://www.youtube.com
 
 Run games by hitting the `Run` button or pressing `Shift+Enter`.
 
-### Infinite loop detection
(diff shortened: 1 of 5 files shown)


grymmy (changes requested):
Please answer a few questions for me:

1. Is this performant? (will it slow down the execution of games?)  How do you know?
2. What is the UX when an infinite loop is detected?
3. Is this more likely to flag as inf loop when your computer is under load? 
4. Where do we intend on documenting this watchdog behavior such that people are not surprised when encountering this?

Separately (I won't demand this), it would be nice to have a few unit tests here proving it works.

src/lib/transform-detect-infinite-loop.ts
```diff
@@ -0,0 +1,89 @@
+// Based on https://repl.it/site/blog/infinite-loops by Amjad Masad.
+
+export default function TransformDetectInfiniteLoop({ types: t }) {
+	const MAX_ITERATIONS = 2000;
```
grymmy, line 4: Suggest moving these to environment variables (or to another constants definition that's higher-level in the project).

grymmy (commented):
Can you also ensure that `debugger` statements continue to work w/ this?

JosiasAurel:
> Please answer a few questions for me:
> 
> 1. Is this performant? (will it slow down the execution of games?)  How do you know?
> 2. What is the UX when an infinite loop is detected?
> 3. Is this more likely to flag as inf loop when your computer is under load?
> 4. Where do we intend on documenting this watchdog behavior such that people are not surprised when encountering this?
> 
> Separately (I won't demand this), it would be nice to have a few unit tests here proving it works.

1. I have tested these changes with a few games in the gallery and there was no noticeable slow down. 
2. This will show up as an error in the editor (see original comment)
3. I'm not sure
4. This will be added directly in the toolkit page in the editor 

image

JosiasAurel:
> Can you also ensure that `debugger` statements continue to work w/ this?

debugger statements still work

Goldenpig12345:
Oh, My Goodness! it's so nice to see you again! How's Anne doing? I
have seen her since she was three!
My doctor has recently said you haven't eaten enough turnips lately. Do you
know if this is true? I really miss you and we need to catch up soon. do
you still have my number? oh, it's okay if you don't I'll just send it to
you. my number is (802)-658-5667. oh ad I heard that your mother was
arrested could you tell me what happened? something about an adult diaper
chase at Walmart? anyway please get back to me soon and maybe we could do
lunch. Love You Lots.

docs/docs.md
```diff
@@ -8,6 +8,23 @@ and sharing them with others. You can watch [this video](https://www.youtube.com
 
 Run games by hitting the `Run` button or pressing `Shift+Enter`.
 
+### Infinite loop detection
```
grymmy, line 11: Suggest cutting this.  Users will experience it intuitively and this probably shouldn't be above the "Getting Help" section.

grymmy (approved):

Issue #1415, opened 2024-01-30
Right click to delete in map and music editor


zachlatta:
Filing this request on behalf of @Al0102.

It'd be nice if you could right click to delete in the map and music editor like you can for the bitmap editor.

leomcelroy:
I support this.

JosiasAurel:
Here are the relevant lines for the map editor
- https://github.com/hackclub/sprig/blob/f650372c4bdf9d23bddbc7d44517ddef078a21ce/src/components/subeditors/map-editor.tsx#L132

And for the sequncer editor 
- https://github.com/hackclub/sprig/blob/f650372c4bdf9d23bddbc7d44517ddef078a21ce/src/components/subeditors/sequencer.tsx#L82

You might want to use the leftDown/rightDown to detect left/right mouse clicks in order to toggle the cell.

GalaxyGamingBoy:
The map editor already has right-click delete.
https://github.com/hackclub/sprig/blob/f650372c4bdf9d23bddbc7d44517ddef078a21ce/src/components/subeditors/map-editor.tsx#L133

PR #1451, opened 2024-02-23
right click to delete note in music editor

GalaxyGamingBoy:
After right-clicking on a sprig sub-editor the cell content should get deleted.
#1415


JosiasAurel (approved):
lgtm

PR #1455, opened 2024-02-27
upgrade esprima to esprima-next and fixing spread operator not working

JosiasAurel:
The `esprima` package has a bug where the spread operator does not work properly on objects (as described in [this issue](https://github.com/jquery/esprima/issues/2034)). There is a fork called [`esprima-next`](https://github.com/node-projects/esprima-next) where this issue is fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
 		"@sendgrid/mail": "^7.7.0",
 		"astro": "2.7.0",
 		"dotenv": "^16.3.1",
-		"esprima": "^4.0.1",
+		"esprima-next": "^6.0.2",
 		"firebase-admin": "^11.10.1",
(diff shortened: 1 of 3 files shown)


grymmy (approved):
@JosiasAurel can u plz add a comment before landing describing what the resulting bug/issue was from a user perspective re: this?

JosiasAurel:
before this change, object spread operator would not work in the Sprig editor.
So, code like this would thrown an error
```
const sampleObject = { name: "kyle", age: 23 }
const inheritedObject = { ...sampleObject, bestCar: "Tesla" }
```

Error: 
```
SyntaxError: Unexpected token ...
    at <game>:131:27
```

PR #1463, opened 2024-02-28
Add extra instruction for setting up a working development environment, add .env.example file

Cosmin-Mare:


README.md
```diff
@@ -118,7 +118,7 @@ cd sprig
 yarn install
 ```
 
-Next, you'll want to give Sprig access to the Firebase credentials you created. Make a `.env` file in the root of the project and enter `FIREBASE_CREDENTIAL=` followed by the base64 string you generated.
+Next, you'll want to give Sprig access to the Firebase credentials you created. Make a `.env` file in the root of the project and enter `FIREBASE_CREDENTIAL=` followed by the base64 string you generated. You will also need to request some extra credentials from the @creds team on Slack to get all the login functionality working.
```
JosiasAurel, line 121: I suggest adding a `.env.example` in the root dir and mentioning all the credentials required in the form `RECAPTCHA_API_KEY=` without explicitly putting the credentials. It will just serve as a guide to tell a contributor what credentials they need to get everything work

JosiasAurel: And here updated the Readme with that credentials need to be setup/updated

JosiasAurel (changes requested):

games/Super_Mario.js
```diff
@@ -0,0 +1,567 @@
+/*
```
JosiasAurel, line 1: please remove this file

games/img/Super_Mario.png
JosiasAurel, line 1: also uncomment this file too

JosiasAurel (changes requested):

PR #1466, opened 2024-02-29
consolidate sprighax repo with sprig repo 

Sheepy3:


grymmy (approved):
suggest making root dir "experiments" and then putting sprig-hax in that

Issue #905, opened 2023-03-04
Add clear button to sound editor


mesaumya:
As I was trying Sprig the first time, it was a little difficult for me to figure out how to clear the engine out until I understood the concept of right-click, also I think it would be easier for the user if there will be an option to clear the engine and the sequencer with one click/button somewhere

alhardwarehyde:
@mesaumya Thank you for your feedback! Do you have a suggestion of where the button should go?

grymmy:
@leomcelroy do you support implementation of this feature?

leomcelroy:
I support this.

PR #1488, opened 2024-03-07
Add Clear Button on Tune Editor

DevIos01:
as mentioned on #905 , i have added clear button which clear the sequence rather than right click each line.

image

image

image


src/components/subeditors/sequencer.tsx
```diff
@@ -136,6 +136,12 @@ export default function SequencerEditor(props: EditorProps) {
 	const preventContextMenu = (event: MouseEvent) => { event.preventDefault(); }
 	window.addEventListener('contextmenu', preventContextMenu)
 
+	const clearSequencer = () => {
```
JosiasAurel, line 139: please can you add comments what each of these reassignments do?

JosiasAurel (changes requested):

JosiasAurel (approved):
perfecto

PR #1514, opened 2024-03-13
improve stability and speed of docker builds and update readme to match

whatwareweb:
This PR is to update the docker image to fix a couple bugs and hopefully make sure it's stable enough to produce builds.
It should also fix some issues people are facing where spade builds but the file is not copied successfully

In addition, the build speed was also decreased.


Sheepy3:
> In addition, the build speed was also decreased.


did you mean increased?

whatwareweb:
Yes that is correct haha

Sheepy3 (changes requested):
could you specify which bugs this pr solves?

whatwareweb:
Yes, when I originally wrote this docker image I didn't know about the security-opt flag, so i implemented a sketchy hack using chcon that is not platform independent. Changing this should fix issues with the docker image accessing outside files.

A bug was fixed where the refresh.sh was run at the wrong time so JerryScript build parameters were not properly set.

Instead of running the jsdev.sh script and waiting 10s and killing it, the commands were directly integrated into the docker build script to increase speed.

There were a couple more minor things fixed as well, not very important though

Sheepy3 (changes requested):
Tested and seems to be working great, but there seem to be merge conflicts as your branch is out of date!

whatwareweb:
Thanks for letting me know, that's fixed now.

I noticed though that the build script was moved to the gardenshed dir, the directory change seems like it is handled in the python script but this means it would not work if the script was run standalone. should i add this to a readme or something to avoid confusion?

whatwareweb:
Fixed more merge conflicts

grymmy (approved):
lgtm, built successfully.  waiting on @Sheepy3 to check on functionality and to compare md5 of build he makes vs. one I made

Sheepy3:
tested app loading correctly, md5 of my built is same as grahams. merging!

PR #1519, opened 2024-03-14
Gallery API Route

GalaxyGamingBoy:
Adds an API route (`/api/gallery`) that supports gallery game fetching.
Supported Query Params:
+ `new`: Filters only new games
+ `query`: Returns all games that match query
+ `tags`: A comma seperated string of tags that games must include

### DIFF
--- a/src/pages/api/gallery.ts
+++ b/src/pages/api/gallery.ts
@@ -1,0 +1,49 @@
+import type { APIRoute } from "astro";
+import {
+	GameMetadata,
+	getGalleryGames,
+} from "../../lib/game-saving/gallery";
+
+export const get: APIRoute = async ({ request }) => {
+	const params = new URL(request.url).searchParams;
+
+	let games = getGalleryGames().map((game) => ({
+		...game,
+		title: game.title.trim(),
+		lowerCaseTitle: game.title.trim().toLowerCase(),
+		lowerCaseAuthor: game.author.toLowerCase(),
+	}));
+
+	games = games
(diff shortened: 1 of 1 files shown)


src/pages/api/gallery.ts
```diff
@@ -0,0 +1,49 @@
...
+		.slice(0, 10)
+		.map((game) => ({ ...game, isNew: true } as GameMetadata))
+		.concat(games.slice(10));
+
+	if (params.has("new")) {
+		games = games.filter((game) => game.isNew);
+	}
+
+	if (params.has("query")) {
+		const query = params.get("query")?.toLowerCase() || "";
```
JosiasAurel, line 28: if `query` is in the params then we don't need to use optional chaining right? We can use `params.get('query')!` (the bang) to tell TypeScript it's not going to be null

src/pages/api/gallery.ts
```diff
@@ -0,0 +1,49 @@
...
+
+		games = games.filter(
+			(game) =>
+				game.lowerCaseTitle.includes(query) ||
+				game.lowerCaseAuthor.includes(query)
+		);
+	}
+
+	if (params.has("tags")) {
+		const tags = params.get("tags")?.split(",") || [];
```
JosiasAurel, line 38: same thing here

JosiasAurel (approved):
This is good. Perhaps you can make those changes?

JosiasAurel (approved):

PR #1555, opened 2024-03-24
TELEPORT FUNCTION

### LINKED ISSUE #1554: teleport(spr, newX, newY)
This feature proposal includes the following function:
```
teleport(sprite, new_x, new_y, intoSolids)
```
The teleport function will teleport a sprite directly to a new location, doing minimal (or none with `intoSolids`) checks to if the sprite can reach the desired point. 
The `intoSolids` parameter is optional, where if it's `true` it will not check if the target is a solid, and if it's `false` or `undefined` (a.k.a. leaving the parameter un-filled), it will check if the target tile contains a solid - and if so, not move the sprite into the new position. 
The function will return a boolean to indicate whether it was successful or not. The only failure conditions should be if the new position is out-of-bounds, or if `intoSolids` is false and the target tile contains a solid.

This solves the problem of portals (see #1282) not working, as with the current implementation of sprite movement in Spade, we can only move sprites one tile at a time, and cannot "force" a sprite to move to a new location.

snoglobe:
i PROMISE this works - i had to do some hacky things to be able to build the firmware due to the migration (hacks that i WON'T commit because they only work locally) but i tested it and it works!

fixes #1554

### DIFF
--- a/firmware/spade/src/shared/sprig_engine/base_engine.c
+++ b/firmware/spade/src/shared/sprig_engine/base_engine.c
@@ -677,6 +677,37 @@ static void map_move(Sprite *s, int big_dx, int big_dy) {
   else             s->dy = moved;
 }
 
+static int map_teleport(Sprite *s, uint16_t new_x, uint16_t new_y, int into_solids) {
+    if (new_x >= state->width) return 0;
+    if (new_y >= state->height) return 0;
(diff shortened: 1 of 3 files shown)


grymmy:
Unfortunately, per #1554, we do not plan on incorporating this functionality into Sprig, sorry...

Issue #1557, opened 2024-03-25
Map editor does not open when clicking on map thingie


grymmy:
**Describe the bug**
After some work in the editor, I observed not being able to open the map editor when clicking on a map variable.

**To Reproduce**
1. Click on a map variable

**Expected behavior**
The map editor opens

**Actual behavior**  
The map editor does not open

**Screenshots**
https://hackclub.slack.com/files/U01FAVARYH1/F06QSUETU59/screen_recording_2024-03-21_at_8.06.06_pm.mov

ryanrudes:
Here is a minimal procedure to reproduce the issue.

Consider the following example:
```js
/*
First time? Check out the tutorial game:
https://sprig.hackclub.com/gallery/getting_started

@title: 
@author: 
@tags: []
@img: ""
@addedOn: 2024-00-00
*/

const player = "p"

setLegend(
  [ player, bitmap`...` ]
)

setSolids([])

let level = 0
const levels = [
  map`
p.
..`
]

setMap(levels[level])

setPushables({
  [ player ]: []
})

onInput("s", () => {
  getFirst(player).y += 1
})

afterInput(() => {
  
})
```

1. Duplicate this line, without adding a separating comma
```js
[ player, bitmap`...` ]
```
2. Click Run. The editor appropriately raises a syntax error:
```
Error: The sprites passed into setLegend each need to be in square brackets, like setLegend([player, bitmap`...`]).
    at setLegend (index.78f90869.js:1:8965)
    at setLegend (index.78f90869.js:3120:12136)
    at eval (index.78f90869.js:14:1)
```
3. Looking at the console events log, the backend also raises this error
```
editor.bde1f794.js:23 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'state')
    at editor.bde1f794.js:23:22531
    at Array.map (<anonymous>)
    at f1 (editor.bde1f794.js:23:22513)
    at editor.bde1f794.js:23:26806
    at p.x (signals.module.5a04f1ba.js:1:1208)
    at p.c (signals.module.5a04f1ba.js:1:3642)
    at $ (signals.module.5a04f1ba.js:1:312)
    at u.set (signals.module.5a04f1ba.js:1:1616)
    at Object.c [as onClick] (editor.bde1f794.js:268:6812)
    at Object.onClick (button.7ef0ffa5.js:8:108)
    at HTMLButtonElement.B (preact.module.3821bb68.js:1:4195)
```

[...]

PR #1563, opened 2024-03-30
Fix for #1557

snoglobe:
fixes the issue of #1557 where the map editor shits itself upon a bad `setLegend()` call due to bad handling of invalid data :)


grymmy (approved):
Does look like a line or two could be more elegant but this solves a major problem! Gj

snoglobe:
tested on vercel - checks out

PR #1580, opened 2024-04-06
IMG Tag

GalaxyGamingBoy:


LucasHT22 (changes requested):
And what happen with people's images?

GalaxyGamingBoy:
per the bug report, the img tag is not used anymore. The image gets dynamically fetched on load

Sheepy3:
this is marked as changes requested, are there actually any changes to be made here @LucasHT22 ?

Cosmin-Mare (approved):
Changes look good to me! On a side note, can't we just remove the img tags altogether and also remove them from the starting code?

GalaxyGamingBoy:
PR is already out for that, pending reviews

Issue #1602, opened 2024-04-21
Spade Docker build doesn't work


recursiveforte:
Building with Docker (on macos, m1) outputs a binary, but uploading games does not work (startup sequence is not read properly and game is never uploaded).

PR #1606, opened 2024-04-22
fix game upload startup sequence

recursiveforte:
reset `c` _before_ reading it instead of after. supports cases where a zero is erroneously read, throwing off the startup sequence.
i'm not really sure why it was broken before (some weird environment edge case?), but this fixes game uploading with docker builds.


recursiveforte:
fixes #1602

grymmy (approved):
lgtm

PR #1611, opened 2024-04-24
Get version number from console & check in web editor on upload.

recursiveforte:
On each upload of a game to the sprig device, the web editor will check if it is running the latest version. If it's running a legacy version (created before this commit & using the old serial interface), the web UI will force the user to update to a later version. If it's running a version that uses the new serial interface, the UI will issue a warning.

I was running into heap overflow issues, so I reduced Jerryscript's heap size from 202kb to 190kb. ~We're not currently able to compile Jerryscript, so I copied the compiled Jerryscript files from an earlier version of spade that had a heap size of 150kb. Jerryscript's internal heap is allocated in the BSS section, which was eating into our normal heap space and explains why we were seeing only 16kb of available heap space.~

The latest version number is declared in `firmware/spade/src/shared/version.json`


firmware/spade/src/rpi/main.c
```diff
@@ -174,9 +175,56 @@ static void rng_init(void) {
...
+                serial_command_indexes[i]++;
+                moved = 1;
+            }
+            if (strlen(serial_commands[i]) == serial_command_indexes[i]) {
+                return i;
+            }
+        }
+        if (!moved) return -1;
+    }
+}
```
JosiasAurel, line 209: please can you explain what this does? Maybe you should also add comments here.

firmware/spade/src/rpi/main.c
```diff
@@ -174,9 +175,56 @@ static void rng_init(void) {
...
+
 // Wait for a game to be uploaded.
 static int load_new_scripts(void) {
-  return upl_stdin_read();
+    int current_command = read_command();
```
JosiasAurel, line 213: ok so `read_command` is supposed to find out what command we're getting to execute?

JosiasAurel (approved):
tested the code and works! it would be nice to have comments in the code especially for `read_command` function

src/lib/upload.ts
```diff
@@ -1,8 +1,11 @@
-import { signal } from '@preact/signals'
+import {signal} from '@preact/signals'
+import {version as latestVersion} from '../../firmware/spade/src/version.json'
 
 export type UploadState = 'IDLE' | 'LOADING' | 'ERROR'
+export type VersionState = 'OK' | 'LEGACY' | 'OLD'
```
grymmy, line 5: could we add commentary here on what each of these mean?

grymmy (approved):

PR #1616, opened 2024-04-25
run gdb client in docker instead of host

recursiveforte:
this gives gdb access to the full source tree for debugging, allowing you to view code as it's being run in GDB


grymmy (approved):
yesssssss

PR #1617, opened 2024-04-25
fix metadata for Box_Guy

recursiveforte:


recursiveforte:
@grymmy can we plz enable the setting in github that requires for all checks to pass before a PR is merged

PR #1620, opened 2024-04-26
Spade multi-game support

recursiveforte:
Add support for the Sprig console to load more than one game.

Games are stored in a metadata section, located after the binary and before the game data section. This metadata section has the capacity to store metadata (game name, size, location) for 32 games, and the version number of the Spade version that last modified this section. Following the metadata, there is a 600kb region set out for games- this is separated into 150 4kb "blocks" which are equal to a flash sector. Games are stored in this region, and can occupy 1-150 slots, so games of varying length can be stored efficiently. This region can be easily expanded in the future, but is left small for now to allow for future features (game persistent state?).

When the console has space to store a game being uploaded but can't store it in a contiguous flash region, Spade will consolidate the storage section to make space for a contiguous region at the end of the section. If the console is out of save slots or metadata slots, it will send a message to the client, and a modal will be displayed on the client.

When this firmware version is installed on a console with an older version and game-saving scheme, it will adopt the existing game stored in flash and create a metadata entry for it, calling it "Legacy Game".

The game engine is no longer stored alongside each game. When running a game, it is now stored only once, in the BSS, and run independently from the game script in js.h. For games stored with legacy versions of the firmware, we read the existing game engine script just for that game- there's a flag in the metadata for whether we should run the engine script independently of the stored game.

Unnecesary JerryScript build files have also been deleted and added to the gitignore.


grymmy:
@recursiveforte do u plan to land this PR?  no biggie if not (or the other way), just triaging... if this PR doesn't serve a useful purpose to u presently, suggest closing it

recursiveforte:
Yeah - working on it today. This is just a dev branch

firmware/spade/src/rpi/main.c
```diff
@@ -265,74 +285,173 @@ int main() {
...
-                     " sprig.hackclub.com \n");
-    render_errorbuf();
-    st7735_fill_start();
-    render(st7735_fill_send);
-    st7735_fill_finish();
+    int games_i = 0;
 
-    load_new_scripts();
-  }
+    int games_len = 8;
```
grymmy, line 308: Please promote this to a define somewhere else with a good descriptor that this is the minimum # of games supported.

firmware/spade/src/rpi/main.c
```diff
@@ -265,74 +285,173 @@ int main() {
...
+        /**
+        * Wait for a keypress to start the game.
+        * This is important so games with e.g. infinite loops don't
+        * brick the device as soon as they start up.
+        */
+
+        if (multicore_fifo_rvalid()) {
+            uint32_t c = multicore_fifo_pop_blocking();
+
+            if (welcome_state == DELETE_CONFIRM) {
```
grymmy, line 337: suggest moving logic for determining welcome state into separate method

firmware/spade/src/rpi/main.c
```diff
@@ -265,74 +285,173 @@ int main() {
...
+                char game_split_lines[] = {
+                        "                    \n"
+                        "                    \n"
+                        "                    \n"
+                        "                    \n"
+                        "                    \n"
+                        "                    \n"
+                };
+
+                for (int i = 0; i*17 < strlen(games[games_i].name); i++) { // 20 - 3 buffer = 17
```
grymmy, line 405: make this more human-readable

firmware/spade/src/rpi/main.c
recursiveforte: can't really do this since it relies on the games list to change index, etc.

firmware/spade/src/rpi/main.c
```diff
@@ -265,74 +320,185 @@ int main() {
...
+                    unsigned int write_length = chars_per_line;
+                    if (strlen(games[games_i].name) - i*write_length < write_length) {
+                        write_length = strlen(games[games_i].name) - i*chars_per_line;
+                    }
+
+                    memcpy(write_dest, game_line, write_length);
+                }
+
+
+                sprintf(errorbuf,
```
grymmy, line 458: errorbuf?  this relates to the screen amirite?

firmware/spade/src/rpi/main.c
```diff
@@ -265,74 +320,185 @@ int main() {
...
+                                  "                    \n"
+                                  "                    \n"
+                                  "                    \n"
+                                  "                    \n"
+                                  "                    \n"
+                                  " sprig.hackclub.com \n"
+                        );
+                break;
+            case RUN_GAME: {
+                // screen is 20 characters wide
```
grymmy, line 406: suggest breaking this out into separate handler method

firmware/spade/src/rpi/main.c
```diff
@@ -265,74 +320,185 @@ int main() {
...
-    render_errorbuf();
-    st7735_fill_start();
-    render(st7735_fill_send);
-    st7735_fill_finish();
+    int games_i = 0;
 
-    load_new_scripts();
-  }
+    int games_len = 0;
+    Game* games = malloc(METADATA_MAX_ENTRIES * sizeof(Game));
```
grymmy, line 344: We might want to in future wrap all malloc's and frees in a #define - not sure if there is any precedent at all but it wouldn't be a bad idea

grymmy (approved):

Issue #1568, opened 2024-04-02
showcase new great games every week


Sheepy3:
![image](https://github.com/hackclub/sprig/assets/109123711/e7cbfde8-deb3-4c01-94f4-8721cd2c2003)
The sprig on the front page showcases a game by a hackclubber from a long time ago, with no attribution/link to the game. I think this would be a good avenue to show off more impressive games that hackclubbers have made, or to feature new games on the front page. credit could also be given by hovering over the game on the 3D sprig.

grymmy:
upvote

DevIos01:
Hey @leomcelroy I know it's low priority but can I get approval for features to be merged [PR#1623](https://github.com/hackclub/sprig/pull/1623) or I should do an alternative way for this task?

grymmy:
@leomcelroy ping on this question from @DevIos01

leomcelroy:
I approve this.

PR #1624, opened 2024-04-27
Comment Console.log so user pc don't overload on background

DevIos01:
while fixing #1568 I saw that some games use loop console.log causing 10k consoles per min making the user's PC work slowly, this would happen on the webpage + anytime someone tries to play games

this comment out so the developer can still uncomment and test locally

as a rule of working smarter not harder, I wrote this code to comment out all in a few mil sec, maybe be used for future usage for something else

```python
import os
import re

folder_path = './games'

for root, _, files in os.walk(folder_path):
    for file in files:
        if file.endswith('.js'):
            file_path = os.path.join(root, file)

            with open(file_path, 'r', encoding='utf-8') as f:
                content = f.readlines()

            updated_content = []
            for line in content:
                updated_line = re.sub(r'(^\s*)console\.log', r'\1// console.log', line)
                updated_content.append(updated_line)

            with open(file_path, 'w', encoding='utf-8') as f:
                f.writelines(updated_content)

print("Completed!")
```


LucasHT22 (approved):
Nice work, LGTM!

PR #1625, opened 2024-04-27
Moving From base64 encode to cdn #cleancode

DevIos01:
after talking with @JosiasAurel, I have now uploaded that base64 image to #cdn as this reduces large chunks of useless code and block/time consuming changing code on the file


JosiasAurel (approved):
nice 👍🏾

Issue #1550, opened 2024-03-22
Add Collaborative Editing to the sprig Editor


Cosmin-Mare:
Collaborative Editing would be a great feature to add to the editor, both for workshops and for people who just want to work together on the same project. For this, I envision the [following UI/UX ](https://www.figma.com/file/a3bUapMcl4tvB769pMiKOC/Untitled?type=design&node-id=0%3A1&mode=design&t=w3XvxEgDUIeWJ6Eb-1) and [a video explaining it](https://drive.google.com/file/d/1qFW1xwP_twwzAc1M7WCMl2UQK6nqD66Q/view?usp=sharing)
the way I envision the UX is: 
1. A host creates a room by going into the editor dropdown, is presented with a popup, and chooses to create a room, for which they’re given a code.
2. Anybody can join the room with that code by doing the same thing, but choosing to join a room instead, inputting the code.
3. The host and co-host should be able to view and edit any participant’s editor by clicking on their name on a collapsable side panel
4. The host and co-host should be able to remove or make co-host anybody in the room
5. The sub editors should be updated In real time, so that people can also draw or make music at the same time as well as collaborate on a project.
6. Anybody would have the possibility to leave the room, and the host would be able to close the room
7. Progress will be saved individually for each person in the room as normal. 
8. if you have a project open in a room, you can’t open it again in another room/separately

leomcelroy:
I approved working on this a few weeks ago. Curious what the current state is?

Cosmin-Mare:
Hey! I have made some demos for some of the functionalities, and am working to soon finish an mvp.

Cosmin-Mare:
This now exists on the beta user flow but implemented in a different way. Will make separate issues for merging this with the main user flow

PR #1650, opened 2024-05-07
Add editor loading skeleton

Cosmin-Mare:
This might not be very useful at the moment, but it will be really useful after implementing some of the other collaborative features (#1550)


grymmy (approved):
lgtm, low risk given not utilized yet

PR #1659, opened 2024-05-09
properly end session after the page is onloaded or the page is closed

JosiasAurel:
Previously after closing the sprig editor window llm chat sessions were not properly ended. 
With these changes, a request will be sent to the server to end that chat session once the window is about to be unloaded


grymmy (approved):
I think this is fine to land but I do wonder if we will still potentially leak sessions because - as I kinda suspected - this event is not guaranteed to fire at all according to the docs:

Screenshot 2024-05-09 at 8 41 11 AM


Some sort of keep-alive heartbeaty thing on backend might be necessary to ensure this never happens (other ideas are welcome)...?  The real test here is someone literally unplugs their computer and it shuts off immediately w/ zero events firing from client->backend - we still shouldn't leak sessions in that scenario.

JosiasAurel:
Chat sessions are closed if no messages are sent after 5 minutes

Issue #1247, opened 2023-11-17
Display warning confirmation when autosaving on a new browser


grymmy:
I propose the new feature:

1. When we send a signal to save off data, we additionally include some ID that is tied to the browser (or session, or whatever makes sense to use).
2. When saving, check to see if the previous save used the same ID.  If it is, save as normal.  If not, display a popup on the client asking the user to confirm they want to potentially overwrite their previous changes.

grymmy:
@leomcelroy there have been multiple reports of data loss due to this sort of scenario.  Please approve/disapprove, or decide on a better/different approach to solve this issue, and assign back to me (or close this, if you find that more appropriate)

grymmy:
@leomcelroy ping on this

leomcelroy:
This seems reasonable. Agree the overriding issue should be addressed.

PR #1679, opened 2024-05-13
 Conflict Confirmation When AutoSaving

DevIos01:
This PR adds support to #1247 after the game mechanic and testing are done (logs will be removed after approval)

Logic:
image

Screenshot:
image


src/components/big-interactive-pages/editor.tsx
```diff
@@ -145,6 +163,38 @@ export default function Editor({ persistenceState, cookies }: EditorProps) {
...
+			}
+		};
+		channel.postMessage({ type: 'SESSION_UPDATE', sessionId });
+		return () => channel.close();
+	}, [sessionId]);
+
+	useEffect(() => {
+
+		const handleUnload = () => {
+			localStorage.removeItem('lastSavedSessionId');
```
grymmy, line 183: why is it important to delete this local session id on unload?  why is this necessary at all?

grymmy (changes requested):

src/components/big-interactive-pages/editor.tsx
```diff
@@ -145,6 +163,38 @@ export default function Editor({ persistenceState, cookies }: EditorProps) {
 	const screenContainer = useRef<HTMLDivElement>(null);
 	const screenControls = useRef<HTMLDivElement>(null);
 
+	const [sessionId] = useState(nanoid());
+
+	useEffect(() => {
```
grymmy, line 168: It is true that all this does detect new sessions?

src/components/big-interactive-pages/editor.tsx
```diff
@@ -66,9 +68,23 @@ let lastSavePromise = Promise.resolve();
 let saveQueueSize = 0;
 export const saveGame = debounce(
 	800,
-	(persistenceState: Signal<PersistenceState>, code: string) => {
+	(persistenceState: Signal<PersistenceState>, code: string, sessionId: string) => {
 		const doSave = async () => {
 			const attemptSaveGame = async () => {
+				const game = (persistenceState.value.kind === 'PERSISTED' && persistenceState.value.game !== 'LOADING') ? persistenceState.value.game : null;
+				if (!game) return false;
+				const lastSavedSessionInfo = localStorage.getItem('lastSavedSessionId');
```
grymmy, line 76: Please modify all usages of the 'lastSavedSessionId' string literal to a constant

src/components/big-interactive-pages/editor.tsx
DevIos01: The **lastSavedSessionId** is removed on page unload to ensure that each session starts fresh, preventing the website from mistakenly displaying an overwrite warning when users return to the editor.

DevIos01: it's replicate how SessionStorage works, it destory on session / tab end, problem with SessionStorage on our case is that it doesn't communicate between tabs, so we need used LocalStorage with this manual remove on unload.

src/components/big-interactive-pages/editor.tsx
```diff
@@ -66,9 +68,23 @@ let lastSavePromise = Promise.resolve();
...
+				const lastSavedData = lastSavedSessionInfo ? JSON.parse(lastSavedSessionInfo) : null;
+				const lastSavedSessionId = lastSavedData ? lastSavedData.sessionId : null;
+				const lastSavedGameId = lastSavedData ? lastSavedData.gameId : null;
+				
+				console.log('Attempting to save. Current session:', sessionId, 'Last saved session:', lastSavedSessionId);
+				
+				if (lastSavedGameId === game.id && lastSavedSessionId !== sessionId) {
+					showSaveConflictModal.value = true;
+					continueSaving.value = false;
+					return false;
```
grymmy, line 86: TBH, I think a better, high-level pattern here would be the following:

```
if(shouldShowConflict(cxt)) {
  showConflict(cxt);
} else {
  if (!save(cxt)) {
    // show error
  }
}
```

Instead, what we're working with here is:
```
if (!save(cxt)) {
  // Figure out why saving failed, it left us a breadcrumb multivariable trail of clues
  if (cxt.continueSaving) { // logic } else if (cxt.showSaveConflictModal) { //logic }....
```

src/components/big-interactive-pages/editor.tsx
DevIos01: yes currently all it does is communicate with session on each tabs aka broadcasting the message to all tabs. (mostly used for testing purposes looking at it now)

grymmy: can this be scoped out and separated to make the logic of how the editor is init'd more concise?

grymmy: i.e. useEffect(logNewSessions)

src/components/big-interactive-pages/editor.tsx
```diff
@@ -147,6 +174,38 @@ export default function Editor({ persistenceState, cookies }: EditorProps) {
 	const screenContainer = useRef<HTMLDivElement>(null);
 	const screenControls = useRef<HTMLDivElement>(null);
 
+	const [sessionId] = useState(nanoid());
+
+	useEffect(() => {
+		console.log('New session started with ID:', sessionId);
```
JosiasAurel, line 180: let's remove the console logs

JosiasAurel (approved):

Issue #1658, opened 2024-05-08
Input box for ai chat disappears 


leomcelroy:
As seen in the video and screenshot below.

It seems linked to resizing the horizontal pane.

![Screenshot 2024-05-08 at 3 20 52 PM](https://github.com/hackclub/sprig/assets/27078897/54f3cd57-551d-4fd5-95f7-96ddd610c997)

https://github.com/hackclub/sprig/assets/27078897/d6a1b830-5bc8-4327-894f-62f4ebca1b03

PR #1682, opened 2024-05-14
prevent help area to resize above output area height

JosiasAurel:
addresses #1658

### DIFF
--- a/src/components/big-interactive-pages/editor.tsx
+++ b/src/components/big-interactive-pages/editor.tsx
@@ -548,7 +548,7 @@ export default function Editor({ persistenceState, cookies }: EditorProps) {
 						/>
 						<div
 							class={styles.helpContainer}
-							style={{ height: realHelpAreaSize.value }}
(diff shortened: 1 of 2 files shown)


src/components/big-interactive-pages/editor.tsx
```diff
@@ -548,7 +548,7 @@ export default function Editor({ persistenceState, cookies }: EditorProps) {
 						/>
 						<div
 							class={styles.helpContainer}
-							style={{ height: realHelpAreaSize.value }}
+							style={{ height: realHelpAreaSize.value, maxHeight: (outputArea.current?.clientHeight! - screenControls.current?.clientHeight!) - 5 }}
```
grymmy, line 551: what is this -5 here?  can it at least be given a name, if its magic?

JosiasAurel: I chose 5 arbitrarily and it should be the screenContainer's height
I've updated it to be `outputAreaHeight - (screenContainerHeight - screenControlsHeight)` so it won't exceed the height of the output area

Issue #1657, opened 2024-05-08
Game window resizes improperly with pane resizing


leomcelroy:
The aspect ratio of the game window should stay fixed to the console aspect ratio. The container should be able to be resized however. 

Below you can clearly see the game sprites getting distorted.

https://github.com/hackclub/sprig/assets/27078897/0f3f757c-d3b4-4b8b-b39a-a5129118f0f1

PR #1684, opened 2024-05-16
make the screen keep it's aspect ratio when resizing help pane

JosiasAurel:
addresses #1657


JosiasAurel:
here's a screen recording of how it looks like 

https://github.com/hackclub/sprig/assets/47951376/226b8b5d-d856-44fa-95ab-a7e0cc20e4f5

leomcelroy:
That doesn't seem quite right. The top of the help section should always be the bottom of the view window.

JosiasAurel:
i looked at how Blot's resizing feature work. i didn't move the screen controls however... is this better? 
https://github.com/hackclub/sprig/assets/47951376/ce73eedb-fa93-4540-8706-046f0adb4153

leomcelroy:
Still not there. The bottom right panel is getting dragged over the top right.

PR #1687, opened 2024-05-20
send a new copy of the code to the server only when the code changes

JosiasAurel:
previously the client chat feature will send 2 requests to the llm api. the first being a copy of the current code and then the chat message sent by the user.

with these changes, we hash the code and only send a new copy of the code to the server when the hash of the current code in the editor is different from the previous one.

### DIFF
--- a/src/components/popups-etc/chat-component.tsx
+++ b/src/components/popups-etc/chat-component.tsx
@@ -18,24 +18,23 @@ const ChatComponent = ({ persistenceState }: ChatProps) => {
 		persistenceState?.value.kind === "IN_MEMORY"
 			? ""
 			: persistenceState?.value.kind === "PERSISTED"
-			? persistenceState?.value.game !== "LOADING"
-				? persistenceState?.value.game.name
-				: ""
-			: persistenceState?.value.name || "";
+				? persistenceState?.value.game !== "LOADING"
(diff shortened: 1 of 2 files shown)


src/components/popups-etc/chat-component.tsx
```diff
@@ -88,7 +90,12 @@ Answer the questions that follow based on this unless new code is provided.`;
 			input.value = "";
 
 			// send code as message to give context to the llm for future questions
-			await sendMessage(newSystemMessage.content);
+			// send new code only if the code has changed since the last one
+			const newCodeHash = await sha256Hash(codeMirror.value?.state.doc.toString()!);
+			if (newCodeHash !== codeHash) {
+			    setCodeHash(newCodeHash);
+					await sendMessage(newSystemMessage.content);
```
grymmy, line 97: do we fail gracefully if this fails...?

grymmy (approved):
plz test carefully and ensure this functions as expected when llm calls fail

src/components/popups-etc/chat-component.tsx
JosiasAurel: it'll only set the code hash if the request is successful. 
there error will also be bubbled-up and handle in the top-level try-catch

PR #1690, opened 2024-05-21
wrap the tags within single quotes

JosiasAurel:
the production deployment failed because the tags were not in valid json format


GalaxyGamingBoy (approved):
LGTM! The formatter also got you, it's not only me it seems

PR #1713, opened 2024-05-25
Make loading skeleton for editor

Cosmin-Mare:
This might not be very useful at the moment, but it will be really useful after implementing some of the other collaborative features (#1550)


GalaxyGamingBoy (approved):
LGTM!

PR #1726, opened 2024-05-30
Add basic syntax highlighting for code responses from Sprig AI

cjdenio:
everybody loves syntax highlighting!

The markdown renderer used here already handles tokenization, so this PR just adds a stylesheet to style tokens based on the editor's default light theme.

Screenshot 2024-05-30 at 11 14 51 AM

There are a few quirks/differences caused by how the markdown renderer works (e.g. function definitions and function calls aren't differentiated, constant declarations are styled while variable declarations aren't), but I think this is better than not having syntax highlighting at all.


faisalsayed10:
thank you so much, this is an epic addition

JosiasAurel (approved):
this looks good 👍🏾

Issue #1396, opened 2024-01-17
Gallery Image Mixup, Non New Games Have 'New' Tag


Omay238:
**Describe the bug**
Both happen in the gallery. Sometimes, games will have the wrong image, and it seems that the games at the top of the list always have the 'new' banner thing. I suspect the image issue is caused by some optimization somewhere, and the 'new' banner issue is probably because it's designed to work with sorting by date, and also it's designed to work with all games (none filtered out)

**To Reproduce**
1. Go to Gallery
2. Search for something (I searched for all of my games)
3. See items with 'new' banner on the top
4. Scroll down, see items with wrong image

**Expected behavior**
Only new games have the new banner, games have the image that is assigned to them, not the image assigned to others.

**Actual behavior**  
Any games at the top after searching have the 'new' banner, and images sometimes show up above the wrong games.

**Screenshots**

<img width="453" alt="Screenshot 2024-01-17 at 4 19 59 PM" src="https://github.com/hackclub/sprig/assets/70179858/6a2002b7-9c9c-41bf-b8b1-1b006b0b89e5">

Omay238:
Expected screen
<img width="453" alt="Screenshot 2024-01-17 at 4 28 57 PM" src="https://github.com/hackclub/sprig/assets/70179858/319b1ff4-b243-47a1-a3ad-37d13739e1cb">

JosiasAurel:
the filtering issue is fixed since #1400 
however I'm unable to reproduce the scenario where the images are incorrectly placed

JosiasAurel:
[...]

PR #1735, opened 2024-06-03
store a copy of gallery images to avoid re-fetching

JosiasAurel:
#1396 

Right now, every time you perform a search the game thumbnail is re-fetched from the server. 

Now game image data is generated at build time and stored in the public folder. These image data are cached through vercel resulting in less resources transferred.

### DIFF
--- a/src/integrations/thumbnail.ts
+++ b/src/integrations/thumbnail.ts
@@ -4,6 +4,7 @@ import fs from 'fs'
 import path from 'path'
 import { dirname } from 'path'
 import { fileURLToPath } from 'url'
+import metrics from '../../metrics'
 
 const __filename = fileURLToPath(import.meta.url)
 const __dirname = dirname(__filename);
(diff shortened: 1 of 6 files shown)


grymmy (approved):
Please test that this doesn't cause issues in worst case scenarios

vercel.json
```diff
@@ -0,0 +1,13 @@
+{
```
grymmy, line 1: This seems very global in nature...  will it have unintended side effects?

src/integrations/thumbnail.ts
```diff
@@ -126,28 +124,22 @@ export const get: APIRoute = async ({ url }) => {
 		}	
 	} catch (error) {
 		// If everything breaks, use a default image
```
grymmy, line 126: tbh... I do wonder if we should instead take the approach of "if anything breaks, break"

src/integrations/thumbnail.ts
```diff
@@ -126,28 +124,22 @@ export const get: APIRoute = async ({ url }) => {
...
-			'Access-Control-Allow-Credentials': 'true',
-			'Access-Control-Allow-Origin': '*',
-			'Access-Control-Allow-Methods': 'GET,OPTIONS',
-			'Access-Control-Allow-Headers': '*',
-			'Cache-Control': 'max-age=86400'
-		}
-	})
+	// write/overwrite image json to public folder
+	let imgFilePath = path.resolve(__dirname, `../../public/${name}.json`);
+	fs.writeFileSync(imgFilePath, JSON.stringify(thumbnail), { encoding: "utf-8" });
```
grymmy, line 137: Please ensure this operation succeeds, and if not have it at least log something.

grymmy (approved):
Please resolve / respond / think about to my other comments before landing.

vercel.json
JosiasAurel: i haven't experienced any unintended side effects using this. i've changed the configuration to only cache those json files still

PR #1757, opened 2024-06-11
Add saving and collab features, not activated

Cosmin-Mare:


src/components/big-interactive-pages/editor.tsx
```diff
@@ -151,6 +152,47 @@ export const saveGame = debounce(
 	}
 );
 
+export async function saveGame2(persistenceState: Signal<PersistenceState>) {
```
grymmy, line 155: suggest better name of this method implying how it functions in a uniqueish way - 2 is noninformational, this "starts the saving process for a game" or something like that

src/components/big-interactive-pages/editor.tsx
```diff
@@ -151,6 +152,47 @@ export const saveGame = debounce(
...
+	const attemptSaveGame = async () => {
+		try {
+			const game =
+				persistenceState.value.kind === "PERSISTED" &&
```
grymmy, line 159: I believe I have observed this same conditional many times in the codebase - should it be wrapped in a convenient method to reduce # of LOC?

src/components/big-interactive-pages/editor.tsx
```diff
@@ -162,14 +204,18 @@ const exitTutorial = (persistenceState: Signal<PersistenceState>, sessionId: str
 			stale: true,
 			cloudSaveState: "SAVING",
 		};
-        saveGame(persistenceState, codeMirror.value!.state.doc.toString(), sessionId);
+		if(isNewSaveStrat.value)
```
grymmy, line 207: The desire here is to be able to enable this feature only for ppl who opt in (you, for instance, to begin with)... the code as I read it here makes me believe that this feature would result in *all* users having this activated - which would be very dangerous and no bueno!

src/components/codemirror.tsx
```diff
@@ -45,22 +58,156 @@ export default function CodeMirror(props: CodeMirrorProps) {
...
-
-	useEffect(() => {
+	useSignalEffect(() => {
+		if(!isNewSaveStrat.value) return
+		if(yProviderAwarenessSignal.value === undefined) return;
+		yProviderAwarenessSignal.value.on("change", () => {
+			yProviderAwarenessSignal.value?.getStates().forEach((state) => {
+				console.log(state)
+				try{
+					if(props.persistenceState === undefined) return;
```
grymmy, line 68: add comment here - what does this imply?

src/components/codemirror.tsx
```diff
@@ -45,22 +58,156 @@ export default function CodeMirror(props: CodeMirrorProps) {
...
+			if(yDoc !== undefined){
+				yDoc.destroy();
+			}
+			if(provider !== undefined){
+				provider.destroy();
+			}
+			yDoc = new Y.Doc();
+			provider = new WebrtcProvider(props.roomState.value.roomId, yDoc, {
+				signaling: [
+					"wss://yjs-signaling-server-5fb6d64b3314.herokuapp.com",
```
grymmy, line 119: this should not be a string literal, but instead in some sort of environment var

src/components/codemirror.tsx
```diff
@@ -45,22 +58,156 @@ export default function CodeMirror(props: CodeMirrorProps) {
...
+				],
+			});
+			//get yjs document from provider
+			let ytext = yDoc.getText("codemirror");
+			const yUndoManager = new Y.UndoManager(ytext);
+
+			yProviderAwarenessSignal.value = provider.awareness
+			console.log(props.persistenceState.peek().session?.user)
+			const isHost = ((props.persistenceState.peek().kind == "PERSISTED" && props.persistenceState.peek().game != "LOADING") && props.persistenceState.peek().session?.user.id === props.persistenceState.peek().game.ownerId)
```
grymmy, line 128: suggest moving this to a separate convenience method for readability of this method

src/components/codemirror.tsx
```diff
@@ -45,22 +58,156 @@ export default function CodeMirror(props: CodeMirrorProps) {
...
+					}, () => onRunShortcutRef.current?.(), yCollabSignal.peek() as Extension),
+					parent: parent.current,
+				})
+
+				setEditorRef(editor);
+				props.onEditorView?.(editor)
+			});
+			yDoc.on("update", () => {
+				if(!props.persistenceState) return;
+				console.log(provider.awareness.getStates());
```
grymmy, line 183: this might be chatty - intentional?

src/components/codemirror.tsx
```diff
@@ -45,22 +58,156 @@ export default function CodeMirror(props: CodeMirrorProps) {
...
+				if (!initialUpdate) return;
+				let participants: RoomParticipant[] = [];
+				provider.awareness.getStates().forEach((state) => {
+					try{
+						participants.push({
+							userEmail: state.user.name,
+							isHost: state.user.host
+						})
+					} catch(e){
+						return
```
grymmy, line 193: what happens if there's an error here?  should this actually blow up instead?

src/components/codemirror.tsx
```diff
@@ -45,22 +58,156 @@ export default function CodeMirror(props: CodeMirrorProps) {
...
+				if(persistenceState.kind === "PERSISTED" && persistenceState.game !== "LOADING"){
+					if(persistenceState.game.ownerId === persistenceState.session?.user.id){
+						saveGame2(props.persistenceState);
+					}
+				}
+				ytext = yDoc.getText("codemirror");
+				initialUpdate = false;
+			});
+		} catch(e){
+			console.log(e)
```
grymmy, line 208: should this be bubbled up?

src/lib/state.ts
```diff
@@ -83,6 +83,24 @@ export type PersistenceState = ({
...
+
+export type RoomParticipant = {
+	userEmail: string
+	isHost: boolean
+}
+
+export type RoomState = {
+	status: RoomStatus
+	roomId: string
+	password: string
```
grymmy, line 100: do we want this in plaintext?  we could keep this encrypted + salt

src/pages/api/games/start-saving.ts
```diff
@@ -0,0 +1,81 @@
+import type { APIRoute } from "astro";
+import {
+	getGame,
+	getSession,
+	updateDocument,
+} from "../../../lib/game-saving/account";
+import { updateEmailListLastModifiedTime } from "../../../lib/game-saving/email";
+import { Timestamp } from "firebase-admin/firestore";
+
+export const post: APIRoute = async ({ request, cookies }) => {
```
grymmy, line 10: Please add a good comment block detailing what this endpoint is and how it is unique

src/pages/api/games/start-saving.ts
```diff
@@ -0,0 +1,81 @@
...
+
+		await updateEmailListLastModifiedTime(session.user, new Date());
+	}
+	const session = await getSession(cookies);
+	if (!session) return new Response("Unauthorized", { status: 401 });
+	if (session.user.id !== game.ownerId)
+		return new Response(`Can't edit a game you don't own`, {
+			status: 403,
+		});
+	// const apiUrl = process.env.SAVING_SERVER_HOST;
```
grymmy, line 53: suggest deleting or putting this in a function whose call is commented

src/pages/api/games/start-saving.ts
```diff
@@ -0,0 +1,81 @@
...
+	// const response = await fetch(`${apiUrl}/listen`, {
+	// 	method: "POST",
+	// 	headers: {
+	// 		"Content-Type": "application/json",
+	// 	},
+	// 	body: JSON.stringify(data),
+	// });
+	// console.log(response)
+
+	updateDocument("games", gameId, {
```
grymmy, line 74: this can fail

grymmy (commented):
Ping me when you have resolved commentary here

src/components/big-interactive-pages/editor.tsx
Cosmin-Mare: I don't think so. It's only 2 conditions, basically 1 line of code

src/components/big-interactive-pages/editor.tsx
```diff
@@ -623,6 +670,7 @@ export default function Editor({ persistenceState, cookies }: EditorProps) {
 								persistenceState.value.tutorial
 							) && (
 								<Help
+									sessionId={sessionId}
```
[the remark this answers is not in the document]
Cosmin-Mare: yes, but it needs its own PR. it solves a bug

src/components/codemirror.tsx
Cosmin-Mare: this implies that something went terribly wrong. it should throw an error instead.

src/components/codemirror.tsx
Cosmin-Mare: it just means that the user won't be counted for. This doesn't happen for users, only for the special saving instance which doesn't have a name or host status, and it should just return and not count it

src/lib/state.ts
Cosmin-Mare: I don't think it's super important to hash it, and it's not yet used.

src/components/big-interactive-pages/editor.tsx
Cosmin-Mare: solved

src/pages/api/games/start-saving.ts
Cosmin-Mare: deleted

src/components/codemirror.tsx
Cosmin-Mare: pretty hard to move the whole thing to a separate method, bc it would require lots of parameters, but I moved a bit of it

src/components/codemirror.tsx
Cosmin-Mare: for now, I'll refresh the page when that happens, because it's very important that that does not raise an error

src/components/codemirror.tsx
```diff
@@ -45,22 +79,135 @@ export default function CodeMirror(props: CodeMirrorProps) {
 		});
 	};
 
-
-	useEffect(() => {
+	useSignalEffect(() => {
```
grymmy, line 82: Plz add high lvl comment for what this signal does

src/components/codemirror.tsx
```diff
@@ -45,22 +79,135 @@ export default function CodeMirror(props: CodeMirrorProps) {
...
+							props.persistenceState.value = {...persistenceState, cloudSaveState: "ERROR"};
+						}
+					}
+				} catch(e){
+					// Do something
+				}	
+			});
+		});
+	});
+	useSignalEffect(() => {
```
grymmy, line 106: plz add high level comment as to what this signal does

grymmy (approved):

PR #1760, opened 2024-06-13
Solve error with saveGame in help.ts (one argument was not passed)

Cosmin-Mare:
the function was not called appropriately, it would've caused an error if the code was run, it's clear without reproducing the bug


grymmy (approved):
lgtm!

JosiasAurel (approved):
lgtm

grymmy:
would be nice to articulate here in a comment what this fixes / why it's necessary

PR #1789, opened 2024-06-19
Add new "vscode" theme.

AndrewCromar:
Hello, I have been wanting a vscode theme for a while and I decided to make it. I believe that I did everything correctly but if there are problems let me know an I will try to fix them.


snoglobe:
@grymmy thoughts? i like it

AndrewCromar:
> @grymmy thoughts? i like it! i would just say that you should make the code editor BG dark, because the contrast is too low otherwise

I was looking through the code to find a variable for the editor background and I couldn't find one, none of the other themes change it either, so I don't think that changing the editor background is currently supported.

grymmy:
@JosiasAurel since you are most aware of the theming feature, can you review this and ensure it's been done properly?

grymmy:
I approve this feature.

JosiasAurel (changes requested):

src/lib/state.ts
```diff
@@ -148,9 +148,18 @@ export const themes: Partial<Record<ThemeType, Theme>> = {
...
 		fgMutedOnAccent: "#6d83ff",
 		background: "#3E29ED",
+	},
+	"vscode": {
+		...baseTheme,
+		accent: "#0078d4",
+		accentDark: "#0062ac",
+		fgMutedOnAccent: "#b6b6b6",
+		background: "#1f1f1f",
+		color: "#cccccc",
```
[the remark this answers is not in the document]
AndrewCromar: fixxed

JosiasAurel (commented):
the code looks good to me. however i would have preferred if the editor looks dark

JosiasAurel (approved):

JosiasAurel:
ping here. @AndrewCromar do you still want to push this forward?

AndrewCromar:
Yes please.

JosiasAurel:
> > @grymmy thoughts? i like it! i would just say that you should make the code editor BG dark, because the contrast is too low otherwise
> 
> I was looking through the code to find a variable for the editor background and I couldn't find one, none of the other themes change it either, so I don't think that changing the editor background is currently supported.

checkout https://github.com/hackclub/sprig/blob/f48c9dbc31a4e84d15089e920934c4f25da9b5c1/src/components/codemirror.tsx#L70 

It's an example of how you can import a new theme and set it when the editor theme changes.

AndrewCromar:
> > > @grymmy thoughts? i like it! i would just say that you should make the code editor BG dark, because the contrast is too low otherwise
> > 
> > 
> > I was looking through the code to find a variable for the editor background and I couldn't find one, none of the other themes change it either, so I don't think that changing the editor background is currently supported.
> 
> checkout
> 
> https://github.com/hackclub/sprig/blob/f48c9dbc31a4e84d15089e920934c4f25da9b5c1/src/components/codemirror.tsx#L70
> 
> It's an example of how you can import a new theme and set it when the editor theme changes.
> 
> For the theme, you could try this library [@uiw/codemirror-theme-vscode](https://www.npmjs.com/package/@uiw/codemirror-theme-vscode)

I've managed to change the background color of the editor (image attached), however now there is another problem with the toolkit where its background does not change.

Also I don't know about you but this theme is starting to look very nice.

image

AndrewCromar:
This comment is a bit long so I broke it down into a few sections.

# Idea
I figured out how to make the help area background color darker, however I think it might be better to leave it light, this is because there are different text colors inside the help area and the all use the accent colors and other vars like that. There would need to be a lot of refactoring to make the background dark with good readability in the help area.

I just don't think it is worth messing with it, the user could always hide the help area if it is to bright for them.

# Example: Dark Theme
I have attached some images to help point out what I am thinking.
You can clearly see that it would be to hard to read.

image
image

# Example: Light Theme
This is what it looks like with it brighter.

image

# Conclusion
In conclusion I am going to commit this as light theme for the help area. If you really think it should be dark then I will go through and attempt to refactor the code to support theming that whole area.

AndrewCromar:
Could someone tell me why Vercel deployment is failing, I would like to fix it but I cannot access the details.

package.json
```diff
@@ -25,12 +26,13 @@
 		"@codemirror/view": "^6.7.3",
 		"@lezer/common": "^1.0.2",
 		"@preact/signals": "^1.1.3",
-		"@sendgrid/mail": "^7.7.0",
+		"@sendgrid/mail": "^8.1.3",
```
JosiasAurel, line 29: why the need to upgrade the sendgrid/mail library?

package.json
```diff
@@ -65,5 +67,6 @@
 		"@types/throttle-debounce": "^5.0.0",
 		"@types/w3c-web-serial": "^1.0.3",
 		"vitest-canvas-mock": "^0.3.3"
-	}
+	},
+	"packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
JosiasAurel, line 71: what does this do?

JosiasAurel (changes requested):
the code looks okay to me. please see my comments

JosiasAurel:
The "your games" section seems to be broken now 
image

package.json
AndrewCromar: Again probably a side effect of my fixing errors. I will look into it.

AndrewCromar:
> The "your games" section seems to be broken now 
> 
> image
> 
> 

That's weird. I will try to fix it.

JosiasAurel:
@AndrewCromar ping 👀

AndrewCromar:
Lol I just lost interest for a while and I have a lot of school work rn. I'll get around to it soon.

grymmy (changes requested):
Sorry for taking a very long time to review this PR.  I attempted to run this locally and it wouldn't render the main sprig landing page...  this needs to be fixed.

Additionally, if we were to merge this, I don't see why we'd need to do large (major v) library upgrades... please revert that part of this change...

grymmy:
I am closing this due to inactivity, albeit a cool feature....

PR #1796, opened 2024-06-20
Fix all TS build errors

Cosmin-Mare:
Went through all TS erorrs and fixed them


grymmy (approved):
LGTM - please test thoroughly in a preview/staging environment before landing.

PR #1836, opened 2024-06-25
Fix editor bug caused by repeatedly replacing CodeMirror instance

### LINKED ISSUE #1834: Opening bitmaps and maps in editor changes game behavior.
**Describe the bug**
Opening bitmaps and maps in editor changes behavior, sometimes breaking games. This occurs when opening the Sprig editor from the home page, not logged in or with a pre-populated game. It appears that this is caused by #1757

**To Reproduce**
1. Go to sprig.hackclub.com
2. Click on 'Editor'
3. Press run
4. Open bitmap editor, or map editor, then repeat step three.

**Video demo:**
https://github.com/hackclub/sprig/assets/84192816/a011a17a-1b79-49fe-a661-4706c3ba2982

recursiveforte:
useSignalEffect re-runs whenever signals within the callback are modified, while useEffect with an empty dependency array will only run once. Changed useSignalEffect -> useEffect for effect that was causing problems.


Cosmin-Mare (changes requested):
this would make the yjs room feature unusable, because the editor needs to change when the signal changes. did you find any issues or bugs that this creates?

Cosmin-Mare:
I've looked at the issue, but you can also accomplish what you want by calling .peek() instead of .value for the persistenceState I think

Cosmin-Mare:
if it's really urgent, and you can't find another way to do this, I approve it and will try to implement it in a different way

recursiveforte:
Sounds good. I'll merge it for now and we can talk more when you get into the office.

PR #1837, opened 2024-06-25
Fix editor bug caused by repeatedly replacing CodeMirror instance

recursiveforte:
Previous fix didn't work.


Cosmin-Mare (approved):
lgtm

PR #1842, opened 2024-06-25
sprig editor log

snoglobe:
new log for sprig editor! shows line numbers on chrome (slash arc) and firefox. safari sucks and doesn't let us retrieve line numbers. supports console.log and console.error!


recursiveforte:
Screenshot 2024-06-26 at 16 25 30

could you remove the dark green vertical border here or make it so that there aren't two color changes between "toolkit" and "get ai help"?

recursiveforte:
The log doesn't scroll back to the bottom when you open it after closing it:

https://github.com/hackclub/sprig/assets/84192816/884a258b-6e36-4803-8b8a-4280166cddd4

Also, I'm getting these errors in the editor when clicking away from the log:

Screenshot 2024-06-26 at 16 30 07

recursiveforte:
Overflow doesn't seem to be handled:

https://github.com/hackclub/sprig/assets/84192816/082c6111-4d76-4f6a-ada3-2372677a4d54

recursiveforte (changes requested):
I'll review the code once the issues I mentioned are addressed!

snoglobe:
the border problems aren't caused by me - those are pre-existing, even when cutting directly off HEAD, and I don't think should be addressed in this PR

src/lib/engine/index.ts
```diff
@@ -61,6 +66,69 @@ export function runGame(code: string, canvas: HTMLCanvasElement, onPageError: (e
...
+						if(browser.indexOf("Chrome") != -1) {
+							// chrome
+							var offset = stack.indexOf("<anonymous>:") + 12;
+							// slice from there until ')'
+							var line = stack.slice(offset, stack.indexOf(")", offset))
+							var strs = line.split(":")
+							return [parseInt(strs[0]) - 2, parseInt(strs[1])]
+						} else {
+							// firefox/etc.
+							var offset = stack.indexOf("Function:") + 9;
```
grymmy, line 92: This seems brittle...

src/lib/engine/index.ts
```diff
@@ -61,6 +66,69 @@ export function runGame(code: string, canvas: HTMLCanvasElement, onPageError: (e
...
+				logInfo.value = [...logInfo.value, {
+					args: args,
+					nums: (()=>{
+						var err = getErrorObject();
+						// get <anonymous>:x:x if on chrome, Function:x:x on firefox
+						// @ts-ignore
+						var stack = err.stack
+						// check browser
+						var browser = navigator.userAgent;
+						if(browser.indexOf("Chrome") != -1) {
```
grymmy, line 83: Suggest breaking this out into several smaller functions and cleaning it up for readability.  I.e. here just have a:

`return extractConsoleErrorNums(...)`

and within that (this is all pseudocode):

```
function extractConsoleErrorNums(...) {
  const browser = getBrowser(userAgent)
  switch(browser) {
    case CHROME: return extractConsoleErrorNumsForChrome(...); break;
    default: return extractConsoleErrorNumsDefault(...); break;
  }
}
```

src/lib/engine/index.ts
```diff
@@ -61,6 +66,69 @@ export function runGame(code: string, canvas: HTMLCanvasElement, onPageError: (e
...
```
grymmy, line 113: See my comment above - this code is copy/pasted and really should be shared between both this and above.

grymmy (changes requested):
Please see my inline comments.

src/lib/engine/index.ts
snoglobe: funnily enough it's actually better than the sprig editor's way of doing it (unfortunately), it is also the only physical possible way this can be done (the JS runtime does not expose line numbers/col ANYWHERE else)

src/lib/engine/index.ts
snoglobe: i meant to do that and forgot, will do

grymmy (approved):

PR #1846, opened 2024-06-26
add dev login feature

recursiveforte:
to use, go to /login with development@hackclub.com as an email and the secret code (DEV_CODE)


grymmy (changes requested):
I think we just want to highjack the existing login flow, detect when a special email is entered, and prompt for a password only in that case...  it should also be the case that if you know the u/p ahead of time, you should be able to programmatically create a new login session

Cosmin-Mare (approved):
lgmt! please change description tho, becasuse it's not just dev mode anymore

PR #1853, opened 2024-06-27
Sprig Case by Auggie

AZ3Dp:
By Auggie:
Please dm me in the slack @Auggie AKA piggy2.0
for any questions 😄


recursiveforte:
Hey! Thanks so much for this submission - I'm really excited about this and will print it at HQ's printers today! Could you also attach the STEP and Fusion source files?

PR #1863, opened 2024-06-28
Fixed problem on landing page with not all content fitting

claynicholson:
after two hours, it ended up only being one line of code lol


Cosmin-Mare (approved):
lgtm!

PR #1939, opened 2024-07-11
Added specific list of tags for games

claynicholson:


GalaxyGamingBoy (approved):
LGTM! However, before merging we should verify all games have correctly mapped tags

Issue #1744, opened 2024-06-07
Modal for naming game upon creation


snoglobe:
Have a modal that forces the user to name their game (obviously keeping ability to rename it in the future) at creation, instead of auto-generating a name. When having multiple games it's easier to pick out one that you named than an auto-generated one, and having this modal would help with that. Just a nicety

mrdarip:
I'm volunteering to work on this issue.

mrdarip:
I'm volunteering to work on this issue.

mrdarip:
I'm volunteering to work on this issue.

DevIos01:
This has already been worked on and merged on #1944 as is live at sprig.hackclub.com so I am closing this!

PR #1944, opened 2024-07-11
Name modal instead of auto generation

ShubhamPatilsd:
dev update, instead of auto generating a name, prompt the user to create one. solves #1744 

image


src/pages/~/new.ts
```diff
@@ -1,10 +0,0 @@
-import type { APIRoute } from 'astro'
```
grymmy, line 1: My suggestion is to leave the existing route and others may have integrated w/ it.  Add yours in an additive/non-conflicting manner.

grymmy (changes requested):
plz resolve my one comment re: the deleted route, and also can u see why the gh action warnings are showing up inline here?

ShubhamPatilsd:
updated

grymmy (approved):

PR #1954, opened 2024-07-12
Added user facing error if an incorrect password is entered

claynicholson:


GalaxyGamingBoy (approved):
LGTM!

grymmy:
@Cosmin-Mare @claynicholson is the intention to land this PR?  i see it now has a conflict, but we should close this one out...

Cosmin-Mare:
I think we should do it with a red text below the text field rather than an alert, so I'm gonna close this

PR #1965, opened 2024-07-15
Gravity Platformer Fix

NotRewd:
A fix for the gravity platformer game.


grymmy:
I didn't test the game content here as it was an update.  i did fix the metadata such that this change wouldn't cause a deployment failure (I think).  Good on you for fixing your game, I encourage you to improve it over time, we will tend to promote ppl/games that get updated, in general.

PR #1974, opened 2024-07-17
Adding unit tests for error handling in the sprig editor

EerierGosling:
I added 9 unit tests that check whether or not the normalizeGameError function is returning the right output - this is the error message that displays on the sprig editor. Right now, these will all fail since the function is returning incorrect output.

### DIFF
--- a/src/lib/engine/error.test.ts
+++ b/src/lib/engine/error.test.ts
@@ -1,0 +1,223 @@
+import { expect, test } from "vitest"
+import {normalizeGameError} from "./error"
+
+// SyntaxErrors (not in evals) seem to go through babel (i.e. have an error.code), so they are not tested here
+
+test('calls a mistyped console.log function (line 2)', () => {
+    const payload = {kind: "runtime", error: new TypeError("console.llog is not a function")}
+    payload.error.stack = `TypeError: console.llog is not a function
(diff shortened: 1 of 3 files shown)


src/lib/engine/error.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+
+test('Infinite recursion (function line 1, return line 2, call line 5)', () => {
+    const payload = {kind: "runtime", error: new RangeError("Maximum call stack size exceeded")}
+
+    payload.error.stack = `RangeError: Maximum call stack size exceeded
+    at recursion (eval at runGame (index.ts:74:14), <anonymous>:4:3)
+    at recursion (eval at runGame (index.ts:74:14), <anonymous>:4:10)
+    at recursion (eval at runGame (index.ts:74:14), <anonymous>:4:10)
+    at recursion (eval at runGame (index.ts:74:14), <anonymous>:4:10)
```
grymmy, line 109: This can be made more concise.

grymmy (approved):
Great job!  Just one nit-pick below, not a huge deal if you don't fix that tbh.  Please follow up this PR with a GH action so these tests auto-execute when ppl send PRs.

PR #1979, opened 2024-07-17
fixing line numbers in sprig editor error handling

EerierGosling:
I shifted the line numbers so they align with the actual errors and added line numbers to infinite recursion and URI errors.

### DIFF
--- a/src/lib/engine/error.ts
+++ b/src/lib/engine/error.ts
@@ -146,7 +146,7 @@ export const normalizeGameError = (gameError: GameError): NormalizedError => {
  * Finds the line and column of innermost error from a stack.
  * This is modified code from V1.
  */
-function findErrorLineCol(stack: string | undefined): [number | null, number | null] {
+function findErrorLineCol(stack: string | undefined, lineOffset: number): [number | null, number | null] {
 	if (!stack) return [null, null]
 
 	let line = null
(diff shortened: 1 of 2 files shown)


src/lib/engine/error.ts
```diff
@@ -147,7 +157,7 @@ function findErrorLineCol(stack: string | undefined): [number | null, number | n
 
 	if (location) {
 		let lineCol = location[1]!.split(":").map(Number)
-		line = lineCol[0]! - 2 - 1
+		line = lineCol[0]! - 2
```
grymmy, line 160: It'd be great eventually to extract this into a #define (or equivilent in this world), and give it an actual name (line offset for <x>)

src/lib/engine/index.ts
```diff
@@ -79,9 +79,14 @@ export function runGame(code: string, canvas: HTMLCanvasElement, onPageError: (e
 		// other errors do not have an error code attached
 		if (!error.code) {
 			const normalizedError = normalizeGameError({ kind: "runtime", error });
-			normalizedError!.line! += 1;
 			return { error: normalizedError, cleanup };
 		} 
+
+		// At least for SyntaxErrors, the type of error shows as "unknown" instead of "SyntaxError" - this replaces that
+		if (error.message.split(":")[0] == 'unknown') {
```
grymmy, line 86: Is there any case where this will result in an uncaught exception?

grymmy (approved):

PR #1992, opened 2024-07-18
sync sub editor changes

### LINKED ISSUE #1929: In the beta user flow, Bitmap to text not working if a line is added above
If you open bitmap in window1 and then you add lines of code in window2 above the bitmap, while having it open in window1, and then modifying the bitmap in window1, it won't change the proper text, as it will remove a character above

JosiasAurel:
also fixes #1929 & #1928


Cosmin-Mare (approved):
lgtm! I tested it locally and it works without breaking anything. TYSM for solving this!!

PR #1999, opened 2024-07-18
Changed login page to have the same theme as the gallery

EerierGosling:
Removed the background and centered the login rectangle.


recursiveforte (approved):
lgtm

PR #2007, opened 2024-07-19
fix audio bug

### LINKED ISSUE #2004: When resetting a game on mobile, audio doesn't reset
**Describe the bug**
When playing a game in the browser of a phone, hitting the reset butting does not cause the audio to reset, so every time its clicked the audio keeps getting layered

**To Reproduce**
1.Go to a game with background music
2. Click on the reset button

**Expected behavior**
The audio should stop and then be played again

**Actual behavior**  
The audio doesn't stop, but is also played again, making it "stack"

BrightTheBackpack:
closes #2004 
i hope this doesn't break anything


JosiasAurel (approved):
lgtm

PR #2008, opened 2024-07-19
Merged styles of create game and login pages

EerierGosling:
I combined the styles of both pages so they both look better and are more consistent.


GalaxyGamingBoy (approved):
LGTM!

PR #2011, opened 2024-07-20
Added V8 of the Sprig Case

AZ3Dp:
Added yet another iteration of my Sprig Case


grymmy:
Thank you for your contributions.  Let's resume this in a couple of months.

PR #2068, opened 2024-08-01
Update to get a sprig page

EerierGosling:
I changed the get a sprig page to fit with the theme for the rest of the website.


recursiveforte:
lgtm

PR #2069, opened 2024-08-01
Move pico-os.uf2 to public/ & update docs

recursiveforte:
Since our deployments are now out of sync with the main branch, we should host the firmware uf2 on the website instead of in GitHub, so there isn't a mismatch between the firmware version and frontend version.


GalaxyGamingBoy (approved):
LGTM!

PR #2144, opened 2024-08-12
Feat: autocomplete

youknowedo:
Added autocomplete to the web editor. Uses the default CodeMirror JavaScript autocomplete with some custom completions for Sprig-specific functions. It also includes tooltips that show the documentation for that function imported from the docs.


youknowedo:
Currently reads the docs, puts it inside an HTMLElements `innerHTML`, and then reads the `innerHTML` again. This is because the regex seemed to have problems with the Unicode characters the "compiled content" would give. Feel free to change it since there must be a more efficient way to do this, I just haven't figured it out yet.

src/lib/codemirror/init.ts
```diff
@@ -1,30 +1,71 @@
...
+	keymap,
+	lineNumbers,
+	rectangularSelection,
+} from "@codemirror/view";
+import { effect, signal } from "@preact/signals";
+import { h, render } from "preact";
+import SearchBox from "../../components/search-box";
+import type { NormalizedError } from "../state";
+import { sprigFunctionCompletion } from "./autocomplete";
+import widgets from "./widgets";
```
JosiasAurel, line 45: It's generally good to avoid formatting changes in a PR as it makes it hard to read the diff. Can you undo the formatting here?

src/lib/codemirror/init.ts
```diff
@@ -44,76 +85,105 @@ export const initialExtensions = (onUpdate: any, onRunShortcut: any, yCollab?: a
...
+							!item.done;
+							item = iter.next()
+						) {
+							count++;
+							if (
+								item.value.from <=
+									update.state.selection.main.from &&
+								item.value.to >= update.state.selection.main.to
+							)
+								index = count;
```
JosiasAurel, line 134: Please undo the formatting changes. It makes it hard to read what has actually changed.

JosiasAurel (changes requested):
I love the way the completion looks. However it might break if the documentation doc changes. 
Now that the Sprig engine is in the sprig repository, it might be better to read from the [exposed engine API](https://github.com/hackclub/sprig/blob/12e832146da281845df051e17543130c4bbd5689/engine/src/base/index.ts#L283)

grymmy:
Closing this due to inactivity.

PR #2063, opened 2024-07-31
Feature - Combine Run Button Into One

DevIos01:
I have fixed the issue we brought on #1971 and now the new UI lets you pick between Run and Run on Device

Before:
{EF5C8B70-573F-4A6A-BB17-81081440F899}

after:
{AE3F8DBD-3D68-43FC-A700-B0F380214945}
{5ED4E1EA-5C24-4B20-AFCB-0BB00BBEB049}

Issue #1971, opened 2024-07-16
Combine 'Run' and 'Run on device' button


grymmy:
The 'run' and 'run on device' button should be consolidated within a drop-down button.

youknowedo:
wouldnt it be better to be able to click on run and have the dropdown only show when you click on the chevron?

PR #2152, opened 2024-08-13
Feat: Run button with Run and dropdown

youknowedo:
Hard to explain. #2063 replaced the `Run` button in the editor with a dropdown where the old `Run on Device` button used to be. This groups the very similar actioned buttons but makes the action of running the game from 1 to 2 clicks (**gasps**). However, it might not seem like much it's annoying and may add up in time. This is why I made the button do the most recent action, and the chevron activates the dropdown. I'm not familiar with Astro and it's been a while since I used React so the code shouldn't be shippable but was merely made to show what I mean.

image
image
Main Run button runs the game

#1971

### DIFF
--- a/src/components/big-interactive-pages/editor.tsx
+++ b/src/components/big-interactive-pages/editor.tsx
@@ -1,38 +1,35 @@
-import styles from "./editor.module.css";
-import CodeMirror from "../codemirror";
-import Navbar from "../navbar-editor";
-import {
-	IoClose,
-	IoPlayCircleOutline,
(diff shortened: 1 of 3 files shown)


recursiveforte (changes requested):
Hey, I love this! Could you make it more clear which action the user is about to trigger when they click on the button by having different sides parts of the button highlighted? Right now, the entire button is highlighted on hover, which doesn't indicate whether clicking will run the game or open the dropdown.

grymmy:
Closing this due to inactivity.

PR #2180, opened 2024-08-18
cast environment variables to valid numbers for comparison/usage

JosiasAurel:
Previously, the Sprig editor will fail to correctly set the lockoutUntil in firebase because the lockout duration was not taken as a valid number.


DevIos01 (approved):
LGTM

PR #2200, opened 2024-08-21
Fix incorrect wording in linux flashing

DillonB07:


GalaxyGamingBoy (approved):
LGTM! Thanks for fixing the docs

PR #2224, opened 2024-08-25
Rename 2 game files because they were named the same as 2 other game files (ignoring case) and it caused problems with git on windows

Cosmin-Mare:


EerierGosling (approved):
lgtm

PR #2235, opened 2024-08-27
Fix name of game being untitled and saving state not showing for room participants

Cosmin-Mare:


src/components/codemirror.tsx
```diff
@@ -86,12 +86,12 @@ export default function CodeMirror(props: CodeMirrorProps) {
 					if(props.persistenceState === undefined) throw new Error("Persistence state is undefined");
 					if(state.saved == "saved"){
 						let persistenceState = props.persistenceState.peek();
-						if(persistenceState.kind === "PERSISTED" && persistenceState.game !== "LOADING"){
+						if(persistenceState.kind === "PERSISTED" && persistenceState.game !== "LOADING" || persistenceState.kind === "COLLAB"){
```
grymmy, line 89: Can we switch all of these string literals to be predefined as constant variables plz?

src/components/navbar-editor.tsx
```diff
@@ -247,11 +247,12 @@ export default function EditorNavbar(props: EditorNavbarProps) {
 					props.roomState?.value.participants.filter((participant) => {
 						if(participant.isHost) return true
 						return false
-					})[0]?.userEmail === props.persistenceState.value.session?.user.email ? props.persistenceState.value.session?.user.email : "???"
+					})[0]?.userEmail === props.persistenceState.value.session?.user.email ? props.persistenceState.value.session?.user.email : "the host"
```
grymmy, line 250: can this string literal be made into a constant and given a good name, like "defaultPersistenceStateEmail" or something?

src/components/navbar-editor.tsx
```diff
@@ -247,11 +247,12 @@ export default function EditorNavbar(props: EditorNavbarProps) {
...
 			}`,
 			SAVING: "Saving...",
 			ERROR: "Error saving to cloud",
 		}[props.persistenceState.value.cloudSaveState];
+		console.log(saveState)
```
grymmy, line 255: will this be overly chatty?

grymmy (changes requested):
see inline comments

JosiasAurel (approved):
lgtm

src/components/big-interactive-pages/editor.tsx
```diff
@@ -520,7 +521,7 @@ export default function Editor({ persistenceState, cookies, roomState }: EditorP
 			}
 		});
 	}, [initialCode]);
-	if(isNewSaveStrat.value && persistenceState.value.kind === "COLLAB" && typeof persistenceState.value.game === 'string')
+	if(isNewSaveStrat.value && persistenceState.value.kind === PersistenceStateKind.COLLAB && typeof persistenceState.value.game === 'string')
```
grymmy, line 524: While very simple and on a single line, this is the type of thing I'd wrap in a function w/ a good name, something that self-documents the decision you're trying to make.  TBH, even referring to this as "**_new_** save strat" isn't really great - it might be more helpful to refer to it as the "backend saving strategy" vs "frontend saving strategy".

I don't expect a change here, just thoughts for the future.

src/components/codemirror.tsx
```diff
@@ -86,12 +87,12 @@ export default function CodeMirror(props: CodeMirrorProps) {
 					if(props.persistenceState === undefined) throw new Error("Persistence state is undefined");
 					if(state.saved == "saved"){
 						let persistenceState = props.persistenceState.peek();
-						if(persistenceState.kind === "PERSISTED" && persistenceState.game !== "LOADING"){
+						if(persistenceState.kind === PersistenceStateKind.PERSISTED && persistenceState.game !== "LOADING" || persistenceState.kind === PersistenceStateKind.COLLAB){
```
grymmy, line 90: I see many copies of this same code - probably should be wrapped in a func.

grymmy (approved):
lgtm, had nits but its fine - plz ensure u test carefully

PR #2239, opened 2024-08-28
slots -> blocks

24c02:
more understandable this way?


recursiveforte:
ty!

src/components/popups-etc/out-of-space.tsx
```diff
@@ -20,7 +20,7 @@ export default function OutOfSpaceModal({ eotMessage }: {eotMessage: Signal<EOTM
 						}
 						{eotMessage.value?.status == "OO_METADATA"
 							&& <>
-							<h2>Your console has hit the limit for number of games stored!</h2>
+							<h2>Your console has hit its maximum number of stored games!</h2>
```
24c02, line 23: this is really good

PR #2281, opened 2024-09-02
Fix bug where room without password would be inaccessible

### LINKED ISSUE #2280: Collaborative room inaccessible without password
If you create a room without a password you won't be able to access it beacuse entering a room requires you type in a password and the field cannot be empty because the button is disabled.

Cosmin-Mare:
fixes #2280


JosiasAurel (approved):
nice!

PR #2286, opened 2024-09-03
Make it so you can just press enter to create a new game instead of manually clicking the button

Cosmin-Mare:


JosiasAurel (approved):
lgtm

PR #2288, opened 2024-09-03
Feat: Stack Changes

GalaxyGamingBoy:
Implements a custom linker script for the memory mappings.  
Currently we place the changes are:
+ Core0 (SCRATCH_X) stack after Core1 (SCRATCH_Y) 
+ Core0 stack size to be alloc be 6KB, Core1 stack size to be alloc 2KB


firmware/spade/src/rpi/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ target_compile_definitions(spade PRIVATE
   # PICO_DEFAULT_UART=0
   # PICO_DEFAULT_UART_TX_PIN=28
   # PICO_DEFAULT_UART_RX_PIN=29
+  JERRY_SYSTEM_ALLOCATOR=1
```
recursiveforte, line 13: why are you switching to the system allocator here? (instead of the default jerryscript BSS block)

recursiveforte (changes requested):

recursiveforte:
pls make sure to test that audio works w/ this pr- if we increase stack size, it might eat into the heap space used for the audio buffer

GalaxyGamingBoy:
> pls make sure to test that audio works w/ this pr- if we increase stack size, it might eat into the heap space used for the audio buffer

Tested a few games today with audio and it appears to work as expected

PR #2322, opened 2024-09-11
fix tune editor going of sync and throwing `Out-of-order effect` error

### LINKED ISSUE #2312: Editing tune editor too fast causes an out-of-order effect error
Steps to reproduce
- Create a new collab room and open open it in another browser window so as to have two different users interacting
- Create a new tune and open the tune editor in both clients
- Make slow edits to the tune and notice how they sync without any problem
- Make a quick edit in the tune editor as if drawing a stroke, for example
- Notice the editor no longer syncs
- Notice there's an out-of-order effect in the console
image

JosiasAurel:
Fixes #2312 

The two useEffect's in `editor-modal.tsx` can be naughty but they help ensure two things
1. If an update comes from the underlying codemirror document, the first effect doesn't run as we're already updating the editor modal from there
2. If an update was originally made from the currently open editor modal, the changes to the codemirror document will not trigger an update to the open editor modal 
 
Both of these help us avoid Out-of-order errors or Cycles

### DIFF
--- a/src/components/codemirror-widgets/open-button.tsx
+++ b/src/components/codemirror-widgets/open-button.tsx
@@ -14,7 +14,7 @@ export default function OpenButton(props: OpenButtonProps) {
 	return (
 		<button
 			class={styles.openButton}
-			onClick={async (event) => {
+			onClick={async () => {
 				if (!codeMirror.value) return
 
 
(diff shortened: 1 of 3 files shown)


Cosmin-Mare (approved):
lgtm! tested and it works as expected in the multi-user editor, and is now syncing no matter the speed of changes

PR #2349, opened 2024-09-15
Plagiarism Checker v2

DevIos01:
# Plagiarism Checker v2

### What's New?

I’ve made some important upgrades to the plagiarism checker. Previously, I used **[Compare50](https://cs50.readthedocs.io/projects/compare50/en/latest/index.html)**, but I’ve now switched to [**MOSS**](https://theory.stanford.edu/~aiken/moss/).

### Key Improvements:

- **Better Plagiarism Detection:** MOSS is more reliable and supports multiple programming languages.
- **Faster Processing:** The runtime has improved significantly! It used to take **8-9 minutes** but now it only takes **1 minute 20 seconds** to complete the check.
- **Smoother Workflow:** I’ve transitioned from a Python-based system to JavaScript, making everything run more smoothly.

#### Before: Runtime with `compare50` (8-9 minutes)

Before - 8-9 minutes runtime

#### After: Runtime with MOSS (1:20 minutes)

After - 1:20 minutes runtime


.github/scripts/plagiarism_check.cjs
```diff
@@ -0,0 +1,236 @@
...
+	const splitFolders = ['.github/split1', '.github/split2'];
+	return { suspectFile, gamesFolder, resultFolder, splitFolders };
+}
+
+async function splitFilesAsync(gamesFolder, splitFolders) {
+	try {
+		log(`Splitting files into ${splitFolders.length} groups...`);
+
+		const jsFiles = await fsPromises.readdir(gamesFolder);
+		const jsFilesFiltered = jsFiles.filter(file => file.endsWith('.js'));
```
JosiasAurel, line 27: could make this into a single line 
```js
const gameFiles = (await fs.readdir("games")).filter(file => file.endsWith(".js"));
```

.github/workflows/check_plagiarism.yml
```diff
@@ -15,37 +15,72 @@ jobs:
...
         run: |
           echo "Pull Request Base SHA: ${{ github.event.pull_request.base.sha }}"
           echo "Pull Request Head SHA: ${{ github.event.pull_request.head.sha }}"
-          git diff --name-only --diff-filter=AM --find-renames --find-copies ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }} | grep 'games/.*\.js$' | xargs
-          js_files=$(git diff --name-only --diff-filter=AM --find-renames --find-copies ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }} | grep 'games/.*\.js$' | xargs)
-          echo "FILES=$js_files" >> $GITHUB_ENV
+
+      - name: Inject MOSS User ID into moss.pl
+        run: |
+          sed -i "s/\$userid=;/\$userid=547529030;/" .github/scripts/moss.pl
```
JosiasAurel, line 64: don't we have to hide the MOSS User ID from the public?

JosiasAurel (approved):
looks okay to me. left some comments

.github/workflows/check_plagiarism.yml
DevIos01: Yes, we should ideally hide the MOSS User ID, but due to GitHub's limitation on using secrets in workflows from forked repositories, we cannot. Secrets are not accessible in this case, so we have to expose the ID temporarily for the workflow to function.

More details on this limitation can be found here: [GitHub: Using Secrets in Workflows](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow).

.github/scripts/moss.pl
```diff
@@ -0,0 +1,354 @@
...
+#
+use IO::Socket;
+
+#
+# As of the date this script was written, the following languages were supported.  This script will work with 
+# languages added later however.  Check the moss website for the full list of supported languages.
+#
+@languages = ("c", "cc", "java", "ml", "pascal", "ada", "lisp", "scheme", "haskell", "fortran", "ascii", "vhdl", "perl", "matlab", "python", "mips", "prolog", "spice", "vb", "csharp", "modula2", "a8086", "javascript", "plsql");
+
+$server = 'moss.stanford.edu';
```
grymmy, line 159: uhhh is this like, making a bunch of api calls to some stanford server....?

grymmy (changes requested):
plz help me understand how this moss component works - are we hitting some random 3rd party server w/ this now?

DevIos01:
> plz help me understand how this moss component works - are we hitting some random 3rd party server w/ this now?

Yes, it's basically sending our game fils to Stanford Server, and it does its own closed source plagarsim and sends us a result 

You can read more at https://theory.stanford.edu/~aiken/moss/

.github/scripts/moss.pl
DevIos01: Yea, we use the code from http://moss.stanford.edu/general/scripts/mossnet

DevIos01: The moss.pl script is the client-side utility provided by the MOSS service. This script helps to send our game files to the MOSS server at Stanford University for analysis.

grymmy: this doesn't seem like it's set up for production workloads...  this is very unusual/weird...

grymmy: heh for instance starting w/ this: 
Screenshot 2024-09-16 at 11 31 07 AM

MichaelK-F:
> > plz help me understand how this moss component works - are we hitting some random 3rd party server w/ this now?
> 
> Yes, it's basically sending our game fils to Stanford Server, and it does its own closed source plagarsim and sends us a result
> 
> You can read more at https://theory.stanford.edu/~aiken/moss/

It might be a good idea to be able to fall back on the old method in the event that the 100 submissions are recieved.

DevIos01:
> > > plz help me understand how this moss component works - are we hitting some random 3rd party server w/ this now?
> > 
> > 
> > Yes, it's basically sending our game fils to Stanford Server, and it does its own closed source plagarsim and sends us a result
> > You can read more at https://theory.stanford.edu/~aiken/moss/
> 
> It might be a good idea to be able to fall back on the old method in the event that the 100 submissions are recieved.

currently, it's 50 games per day it can do plagiarism on, and with our current setup we don't need more than that and it is currently enough with extra room in space, i am working on adding a rotation key for this fallback so  it uses another  user id making it 100 games per day

grymmy:
I am closing this in preference to the new plag checker.

PR #2376, opened 2024-09-21
Update "Report a bug" button to be less eyecatching (#1972)

### LINKED ISSUE #1972: Make 'report a bug' slightly less eye-catching
The large red button attracts more attention than it deserves.  Make it slightly more subtle.

Hex-4:
I've changed the "Report a bug" button in the editor to the same color as the Share/Save button in the navbar. It still becomes red when you hover. Should fix #1972.

note - this is my first dev PR here so something will probably be off. like for example it shows 10k additions and 7k deletions which i don't think should be happening.


grymmy (approved):
Seems pretty reasonable to me.

PR #2399, opened 2024-09-30
Fix bug where saving state wouldn't be reported accurately to room participants + refactor editor return statement (does the same thing, but looks better)

### LINKED ISSUE #2369: Saving state not accurately reported to room participant that is not the host
To reproduce:
1. Make a shareable room on the beta editor
2. Open that room in a different browser with a different account
3. Make changes and observe that it permanently says saved to host, instead of saying saving...

Cosmin-Mare:
fixes #2369

### DIFF
--- a/src/components/big-interactive-pages/editor.tsx
+++ b/src/components/big-interactive-pages/editor.tsx
@@ -522,236 +522,240 @@ export default function Editor({ persistenceState, cookies, roomState }: EditorP
 			}
 		});
 	}, [initialCode]);
-	if(isNewSaveStrat.value && persistenceState.value.kind === PersistenceStateKind.COLLAB && typeof persistenceState.value.game === 'string')
-		return (
-			<RoomPasswordPopup persistenceState={persistenceState} />
-		)
-	else
-		return (
-			<div class={styles.page}>
-				<Navbar persistenceState={persistenceState} roomState={roomState}/>
 
-				<div class={styles.pageMain}>
-					<div className={styles.codeContainer}>
-						<CodeMirror
-							persistenceState={persistenceState}
-							roomState={roomState}
-							class={styles.code}
-							initialCode={initialCode}
-							onEditorView={(editor) => {
-								codeMirror.value = editor;
-								setTimeout(() => foldAllTemplateLiterals(), 100); // Fold after the document is parsed (gross)
-							}}
-							onRunShortcut={onRun}
-							onCodeChange={() => {
+	return (
+		isNewSaveStrat.value && persistenceState.value.kind === PersistenceStateKind.COLLAB && typeof persistenceState.value.game === 'string' 
+		? 
+			(<>
+				<RoomPasswordPopup persistenceState={persistenceState} />
+			</>) 
+		:
+			(
+		<div class={styles.page}>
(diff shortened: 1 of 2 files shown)


src/components/big-interactive-pages/editor.tsx
```diff
@@ -522,236 +522,240 @@ export default function Editor({ persistenceState, cookies, roomState }: EditorP
...
+			<div class={styles.pageMain}>
+				<div className={styles.codeContainer}>
+					<CodeMirror
+						persistenceState={persistenceState}
+						roomState={roomState}
+						class={styles.code}
+						initialCode={initialCode}
+						onEditorView={(editor) => {
+							codeMirror.value = editor;
+							setTimeout(() => foldAllTemplateLiterals(), 100); // Fold after the document is parsed (gross)
```
grymmy, line 546: What necessitates folding the template literals to begin with?  I don't argue against needing to, just looking to understand (it's probably best to state this explicitly in the comment, tbh)

src/components/big-interactive-pages/editor.tsx
```diff
@@ -522,236 +522,240 @@ export default function Editor({ persistenceState, cookies, roomState }: EditorP
...
+								);
+							}}
+						/>
+						<div
+							class={styles.helpContainer}
+							// style={{ height: realHelpAreaSize.value, maxHeight: outputArea.current?.clientHeight! - (screenControls.current?.clientHeight! + screenContainer.current?.clientHeight!) }}
+							style={{ height: realHelpAreaSize.value, maxHeight: outputArea.current?.clientHeight! - (screenContainer.current?.clientHeight! - screenControls.current?.clientHeight!) }}
+						>
+							{!(
+								(persistenceState.value.kind === PersistenceStateKind.SHARED ||
```
grymmy, line 697: It really does seem like there are a multitude of these checks which probably could be collapsed into one or several utility methods to make the code more readable and less prone to making a copy/paste or logic error.

grymmy (changes requested):
plz at least add requested commentary - bonus points if u could expend some effort to clean up duplicate persistentState code

src/components/big-interactive-pages/editor.tsx
Cosmin-Mare: that was done by someone else b4 my changes. it's probabyl because every template literal starts up as unfolded by default

src/components/big-interactive-pages/editor.tsx
grymmy: okay but actually plz do follow up on it

src/components/big-interactive-pages/editor.tsx
grymmy: okay - just make sure any refactorings you might do don't break any of this stuff :)

grymmy (approved):

src/components/big-interactive-pages/editor.tsx
Cosmin-Mare: yup, most of the highlighted lines were just tabs

src/components/big-interactive-pages/editor.tsx
Cosmin-Mare: 🫡

PR #2401, opened 2024-10-01
fold template literals after update has been propagated in editor

### LINKED ISSUE #2368: Bitmaps/maps unfold for everyone when someone edits them in one of the instances

JosiasAurel:
resolves #2368 

when a change is propagated to bitmaps/map/tune, their template literals get unfolded and the change applied. however, they're not closed back. this change forces the template literals to fold after these changes have been applied to the editor.

### DIFF
--- a/src/components/big-interactive-pages/editor.tsx
+++ b/src/components/big-interactive-pages/editor.tsx
@@ -88,6 +88,14 @@ const minHelpAreaHeight = 32;
 let defaultHelpAreaHeight = 350;
 const helpAreaHeightMargin = 0; // The margin between the screen and help area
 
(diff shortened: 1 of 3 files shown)


Cosmin-Mare (changes requested):
I don't think we should collapse all template literals when someone edits one of them

grymmy:
Do we plan on this PR surviving and moving forward, or do we kill it with 🔥 ?

Cosmin-Mare:
Afaik @JosiasAurel is working on finding a way to only collapse the bitmap that has been edited, so no🔥yet

Cosmin-Mare (approved):
lgtm, tested and it works perfectly!

PR #2416, opened 2024-10-04
Fix on Game Publishing: Optional Thumbnails and Cookie Policy Update

DevIos01:
This PR contains few changes to add polish on Publish To Github feature:

1. **Cookie Policy:** Set SameSite policy for session cookies to 'Lax' for better security while maintaining usability in authentication flows.
2. **Optional Thumbnails:** Thumbnails are now optional during publishing, with improved handling.


src/lib/game-saving/account.ts
```diff
@@ -233,7 +233,7 @@ export const makeOrUpdateSession = async (cookies: AstroCookies, userId: string,
 		path: '/',
 		maxAge: 60 * 60 * 24 * 365,
 		httpOnly: true,
-		sameSite: 'strict'
+		sameSite: 'lax'
```
grymmy, line 236: What does this change?

grymmy (approved):
Generally this looks good.

src/lib/game-saving/account.ts
DevIos01: By changing from "Strict" to "Lax" we make sure the session token is readable for "new user" so they don't face the "No Active Session" error during publishing process.

while still keeping things secure, "strict" blocks any session and only allowing the main site to do stuff, meanwhile "lax" as same security like strict so it still is prevent from CSRF attack but also allow it to share the session to popup

grymmy (approved):
Would like to see more inline comments here.  Thank you for adding instrumentation that'll help us understand if/when this breaks.  Shocked (in kind of a good way) to see u are actually trying to support the 'update the PR' flow w/ this, wow.

src/components/navbar-editor.tsx
```diff
@@ -471,56 +525,119 @@ export default function EditorNavbar(props: EditorNavbarProps) {
...
+				if (imageBase64) {
+					imageBlobSha = await createBlobForImage(accessToken, forkedRepo.owner.login, forkedRepo.name, imageBase64.split(',')[1]);
+				}
+			} catch (error) {
+				reportMetric("github_publish.failure.image_blob");
+				throw new Error("Failed to create image blob: " + (error instanceof Error ? error.message : String(error)));
+			}
 
-			publishSuccess.value = true;
+			const sanitizedGameTitle = gameTitle.replace(/\s+/g, '-');
```
grymmy, line 576: Is this needed?

DevIos01: We are just doing this for filename, not the game title restrictions

src/pages/api/auth/github/callback.ts
```diff
@@ -41,15 +42,11 @@ export const get: APIRoute = async ({ request, cookies }) => {
...
-			console.error("No active session found");
-			return new Response(
-				'<script>window.opener.postMessage({ status: "error", message: "No active session" }, "*"); window.close();</script>',
-				{
-					headers: { "Content-Type": "text/html" },
-				}
-			);
+			// If no active session, create one
+			console.warn("No active session found, creating a new session.");
+			sessionInfo = await makeOrUpdateSession(cookies, userId, "code");
```
recursiveforte, line 49: you shouldn't be able to call this callback if you're not logged in! this code (makeOrUpdateSession) will log the person in without validating their authentication

src/lib/game-saving/github.ts
```diff
@@ -411,4 +411,13 @@ export async function createBlobForImage(
 
 	const data = await handleResponse(response); // Return the blob SHA
 	return data.sha;
-}
\ No newline at end of file
+}
+
+export const synchronizeForkWithUpstream = async (accessToken: string, forkOwner: string, forkRepo: string) => {
```
recursiveforte, line 416: this is dead code & should be removed

recursiveforte (changes requested):

recursiveforte:
Screenshot 2024-10-18 at 10 53 08
this doesn't work for me- i get this error

recursiveforte:
todo: check in one week to see what happens when github cookies expire (this might've been an issue?)

recursiveforte (approved):

PR #2420, opened 2024-10-05
Fix bug where room participants wouldn't update

### LINKED ISSUE #2419: Participant count of room doesn't increase

Cosmin-Mare:
fixes #2419

### DIFF
--- a/src/components/codemirror.tsx
+++ b/src/components/codemirror.tsx
@@ -181,6 +181,21 @@ export default function CodeMirror(props: CodeMirrorProps) {
 				if(props.roomState)
 					props.roomState.value = { ...props.roomState?.value, connectionStatus: ConnectionStatus.CONNECTED };
 		});
+			provider.awareness.on("update", () => {
+				let participants: RoomParticipant[] = [];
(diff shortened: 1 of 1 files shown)


grymmy (approved):
Would appreciate addition of a singular comment to explain what you're doing here in layperson's terms.

src/components/codemirror.tsx
```diff
@@ -181,6 +181,21 @@ export default function CodeMirror(props: CodeMirrorProps) {
...
+			provider.awareness.on("update", () => {
+				let participants: RoomParticipant[] = [];
+				provider.awareness.getStates().forEach((state) => {
+					try{
+						participants.push({
+							userEmail: state.user.name,
+							isHost: state.user.host
+						})
+					} catch(e){
+						return
```
grymmy, line 193: is it OK for this to be very quietly failing w/o any logging?

Cosmin-Mare: yup. bc that means that the participant doesn't have an email, therefore it's the saving instance and we don't count that towards the participants

PR #2430, opened 2024-10-09
new board for new screen

recursiveforte:
Co-authored-by: Nora <nora@hackclub.com>


grymmy (approved):
Per our discussion might be good to tack a git tag on here such that we could go back easily if needed.

PR #2534, opened 2024-11-01
Add error for when user enters wrong room password

### LINKED ISSUE #1934: Invalid room password does not give an error
There is a silent error currently when you enter the wrong password for a room.  It should prompt the user that they didn't use the right credentials somehow.

Cosmin-Mare:
fixes #1934


grymmy (approved):
lgtm

PR #2574, opened 2024-11-08
[Sprig App] Save The Seas

RyanBran888:
### Author name
Author: RyanBran888

### About your game

**What is your game about?**
The player is a fish, and has to swim to the ocean to collect the litter and save the world!

wasd to move the fish


gusruben:
Nice game, I like the theming! Thanks for your submission :)

PR #2599, opened 2024-11-13
Improve tuning of notes in music engine

whatwareweb:
This change to the music engine will vastly improve the tuning of notes in the music engine of the spade firmware.

The tuning notes issue is due to rounding errors within piano.c. This changes the music functions to use floating point math for higher accuracy.

The changes to the gitignore is to avoid uploading dev builds of firmware.uf2 file to github, the export uf2 filename was changed from the docker script but wasn't updated in the gitignore.


whatwareweb:
This new commit changes the shape of the waves to improve tuning as well

grymmy:
I cannot justify landing this without additional testing, and I cannot do this testing myself.  Please reopen this PR if you're willing to provide proof that this functions without issue (including on device)

whatwareweb:
I don't think I have permission to reopen this pull request, but I'll build spade and send some before/after audio captures in a little while.

whatwareweb:
I've recorded some before and after audio clips, but Github doesn't like my wav files so i'll ping you on Slack.

grymmy (approved):

PR #2604, opened 2024-11-14
[Sprig App] Sprong

Licnex:
### Author name
Author: Licnex

### About your game

**What is your game about?**
Its classic pong where you have to protect your side from the goal. There are two game modes, single player and multiplayer.

press w and s for player 1 (the one on the left side) and i and k for the right hand side


alx-alexpark:
Duplicate of #2605

PR #2606, opened 2024-11-14
[Sprig App] Save the balloon

hrit09:
### Author name
Author: hrit09

### About your game

**What is your game about?**
Save the balloon as long as you can

Press A to move left Press D to move right


alx-alexpark (changes requested):
Cool game!

I am slightly worried by the fact that your game might not reach the required 1 minute of gameplay, but if you make it so you can press a key to restart the game, I'll accept it :)

PR #2642, opened 2024-11-25
[Sprig App] Infinite Platformer

EnterpriseGoose:
### Author name
Author: EnterpriseGoose

### About your game

**What is your game about?**
Try to last as long as you can in this platformer that goes on forever! As you go it will get more difficult, with wider and wider jumps. See if you can get the highest score!

The controls can be a bit tricky. (a) and (d) are used to move left and right and (w) is used to jump. You need to restart pressing the (a) or (d) keys after you jump, and it might take a bit for your keyboard to start repeating the character after you hold it down. If you can, I would suggest changing the keyboard 'repeat delay' and 'repeat rate' to be as short and as fast as possible, respectively. Good luck!


Cosmin-Mare (approved):
Thanks for your submission! I like playing your game, but couldn't get above 35k score:)) the platforms are way too fast for me. But I'm approving your game, and you'll soon be able to see it in the gallery!

Hack Clubbers would love to see this! Just post a link and 1-2 sentences talking about your game to our [#ship](https://hackclub.slack.com/archives/C0M8PUPU6) [channel](https://hackclub.com/slack/):)

PR #2664, opened 2024-11-29
[Sprig App] EMOJI RUSH CLUB tap A or D

ZahriyehNoah:
### Author name
Author: ZahriyehNoah

### About your game

**What is your game about?**
Emoji Rush Club is a web interactive game based on mood and creative ideas. You are expected to consume Challenger objects. Enjoy

Enter A to move Left or Enter D to move right


Cosmin-Mare (changes requested):
hey! it looks like there's some sort of a mixup here, because the file u submitted is the apple drop game, but your description doesn't match with it, so I'm assuming u uploaded or downloaded the wrong file. could you please make another PR/change the files here to your game?

PR #2667, opened 2024-11-29
[Sprig App] the untitled game ultra

Rayaan4819:
### Author name
Author: Rayaan4819

### About your game

**What is your game about?**
its a game that is purly text based (for now), it is a choice game, where as you solve riddles you unlock new paths which therefore unlock new endings (7 endingd for now)

you will get 3 options, choose any three options to get diffrent paths


gusruben (changes requested):
Thanks for your submission! It seems there's a bug on line 123 though that stops the program from running, would you mind checking that out?

Also, from reading the code it looks like this might use browser functions like `prompt` and `alert`. I like the thinking outside the box, but to accept the game it would need to be built in the Sprig engine! If you modified/rewrote the game to use Sprig I'd be happy to approve it!

Rayaan4819:
Thanks for the help, Yea it did have browser elements and its kind of just
an idea i got over the weekend and will definitely go more in detail once
my exams are over, Thanks fp=or your time

grymmy:
@Rayaan4819 do you plan on moving forward with this PR?

grymmy:

PR #2705, opened 2024-12-13
[Sprig App] Genius Hour

GangleTADC:
### Author name
Author: GangleTADC

### About your game

**What is your game about?**
This is a TADC inspired game...

WASD


gusruben (changes requested):
Thanks for your submission, I like the pixel art! I only have one note-- when I finish the game I get an error: `TypeError: can't access property "x", playerSprite is undefined`. Other than this the game is great, so if you fix this I would be happy to approve it!

grymmy:
@GangleTADC do you plan on moving forward w/ this?

grymmy:

PR #2712, opened 2024-12-14
[Sprig App] 2Ptag

Shadow8928:
### Author name
Author: Shadow8928

### About your game

**What is your game about?**
2 players player 1 gets w,a,s,d keys and player 2 gets i,j,k,l keys first person to touch the other win due to issues in the game structure if other wise player 2 can move 2 blocks in one direction while player 1 moves one block the total goal is to touch your opponents player first.

Player 1 moves up down left right one block with W,S,A,D
Player 2 move up down left right two blocks with I,K,J,L


gusruben (changes requested):
Thanks for your submission! Unfortunately, I'm worried it's not complex enough to reach the required 1 minute of playtime, and there's no clear finish condition-- do you think you could add some kind of end screen and/or timer? If you did I'd be happy to re-review!

Shadow8928:
I have added an end game game system to indicate the winner and removed the turn based system. for the issue of it being a minute, the game is relatively simply but time consuming as it is a challenge of skill so time can be as short or long as possible based on skill of players so I am confident it won't be a problem as long as both players understand the game however if it is still a concern about time I could modify it to last longer

grymmy:
Congratz - if you're a teenager, and haven't already received a Sprig device of your own, please fill out this form!  Additionally, feel free to join Hack Club's Slack and share you creation in the #sprig channel!

PR #2715, opened 2024-12-17
warn against using backticks in sprite keys

### LINKED ISSUE #2297: Map editor generates invalid syntax with sprites with backtick key
**Describe the bug**
When a sprite with a backtick (`` ` ``) as its key is placed using the map editor, the editor places the backtick unescaped, ending the string literal causing invalid syntax which is not obvious until closing the map editor.

**To Reproduce**
1. Add a sprite with a backtick as a key to the legend
2. Add the sprite to a map in the map editor

**Expected behavior**
The sprite is placed on the map without breaking syntax, possibly through escaping the backtick.

**Actual behavior**  
The backtick is placed in string and terminates the string early.

**Screenshots**

https://github.com/user-attachments/assets/efd36695-f586-497f-a4b7-77726b619ff2

Since escaping characters means that 1 character may not always = 1 tile, it might be better to just prevent using backtick for a sprite key.

(A separate issue where the map clicked on to edit is not the one that ends up being edited is also exhibited in the video but is not the subject of this issue.)

JosiasAurel:
Fixes #2297 
This PR adds a babel transform to throw an error when there's a backtick in a string. This is to prevent users from including backticks in sprite names which will break on maps. 

It also adds an effect in the editor that will perform checks and bubble errors every two seconds.


src/components/big-interactive-pages/editor.tsx
```diff
@@ -36,6 +36,15 @@ import { PersistenceStateKind } from "../../lib/state";
 
 let screenShakeSignal: Signal<number> | null = null;
 
+const performSyntaxCheck = () => {
```
grymmy, line 39: how often will this be run?

src/lib/engine/index.ts
```diff
@@ -112,19 +132,11 @@ export function runGame(code: string, canvas: HTMLCanvasElement, onPageError: (e
...
-        });
-        logInfo.value = [];
-        const fn = new Function(...engineAPIKeys, transformResult.code!);
-        fn(...Object.values(api));
-        return { error: null, cleanup };
-    } catch (error: any) {
-        onPageError(normalizeGameError({ kind: "runtime", error }));
-        return { error: normalizeGameError({ kind: "runtime", error }), cleanup };
-    }
+	return { error: transformAndThrowErrors(code, engineAPIKeys, (transformedCode) => {
```
grymmy, line 135: how confident are we in this change?  how carefully have you tested it, correctness seems very important here...

grymmy (commented):
plz respond to q's and will take another look afterwards

JosiasAurel:
I've tested the change in editor and added two more tests in the engine. 
image

grymmy (approved):

PR #2622, opened 2024-11-20
Add files via upload

SigmaSkibidiNumbers:
James Wilson

Johamez

My game is a short puzzle game where you push rocks out of the way to get access to a switch which will open up a wall
and give you access to the next level

W A S D - Movement
J - Toggle the switch
K - Reset to level 1 (for when you finish)

- [x ] The game was made using the [Sprig editor](https://sprig.hackclub.com/editor).
- [x ] The game is placed in the in the [`/games` directory](https://github.com/hackclub/sprig/tree/main/games).
- [x ] The code is significantly different from all other games in the [Sprig gallery](https://sprig.hackclub.com/gallery) (except for games labeled "demo").
- [x ] The game runs without errors. 
- [x ] The name of the file/game contains only alphanumeric characters, `-`s, or `_`s.
- [x ] The game name is not the same as the others from [gallery](https://sprig.hackclub.com/gallery)

PR #2722, opened 2024-12-17
[Sprig App] PuzzleMania

SigmaSkibidiNumbers:
### Author name
Author: Johamez Wilzoney

### About your game

**What is your game about?**
A Puzzle game about flipping switches and pushing rocks

There are 5 bonus levels unlocked by a key on a secret level (the portal to said secret level is on lv 2)

"W A S D" to move
"I" to Mute music
"J" to interact with things
"K" to reset level
"L" for level select


gusruben (changes requested):
Thanks for your submission, I like it a lot! Unfortunately I seem to be getting an error when flipping the switch in the second level 4: `TypeError: can't access property "x", getFirst(...) is undefined`. If you fix this I'll be happy to re-review!

Also, a very optional suggestion: some indicators for which levels are finished would help a lot with skill issues like mine :)

SigmaSkibidiNumbers:
I fixed the issue and changed the fake wall texture slightly to make it a
bit easier to spot, I didnt intend for the error on Level B4, however the
lever was meant to be a trick that forces you to restart the level as it
deletes the portal you need to complete the level.

grymmy:
@SigmaSkibidiNumbers you stated you fixed the issue that @gusruben pointed out here - can you please link to the commit where this was done?  am I missing something?

gusruben:
It seems like this PR is a duplicate of #2622, which is more up-to-date-- check for feedback there!

PR #2723, opened 2024-12-17
[Sprig App] CK

jasonG4:
### Author name
Author: jasonG4

### About your game

**What is your game about?**
get all coins in the levels

how to play 
W up
S down
A left 
D right


gusruben (changes requested):
Thanks for your submission! Unfortunately, I'm worried it doesn't reach the required 1 minute of playtime. Do you think you could add some more levels and maybe an end screen? If you did I would be happy to re-review your game!

grymmy:

PR #2736, opened 2024-12-19
[Sprig App] Under the Sea

Finn6360:
### Author name
Author: Finn6360

### About your game

**What is your game about?**
The goal is to catch all the fish so that the octopus can eat. Colliding with the fish counts as catching them. However, don't collide with the boats or it will be game over.

WASD for movement.

### DIFF
--- a/games/Under-the-Sea.js
+++ b/games/Under-the-Sea.js
@@ -1,3 +1,10 @@
+/*
+@title: Under the Sea
+@author: Finn6360
+@tags: []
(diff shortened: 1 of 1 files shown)


Cosmin-Mare (changes requested):
Hey! Your game appears to still be a WIP because neither fish nor obstacles spawn. Please open another PR or update this one when you have completed it and we will review it again! Thank you for your submission!

grymmy:

PR #2743, opened 2024-12-20
[Sprig App] save the goat!!!

Cookiesgobrr:
### Author name
Author: Cookiesgobrr

### About your game

**What is your game about?**
classic move your character or you die ahh game

A and D to navigate da goat. Don't get thunked on.


grymmy:

PR #2761, opened 2024-12-25
Fixing Mailman.js

EerierGosling:
A function was defined twice which made it unplayable.


Cosmin-Mare (approved):
maybe just delete it? I think that would be better

EerierGosling:
@Oasx if you'd like to fix it yourself that'd be great since I wasn't sure which function was the right one to comment out

PR #2791, opened 2025-01-13
Add a jam button

gusruben:
image

Added at the request of @JaredSene -- let me know if it's too jarring or should be moved somewhere else


grymmy:
Very reasonable, thanks!

PR #2815, opened 2025-01-22
[Sprig App] subway surfish

scottwellcorn:
### Author name
Author: scottwellcorn

### About your game

**What is your game about?**
it a endless game of the same level which gets faster and faster with every time you complete it

your constantly moving forwards and you can only move up(w) and down(s)


Cosmin-Mare (changes requested):
Hey! Thanks for your submission. Your game is not really playable in the sense that it's too small and it's confusing and doesn't meet the 1m playtime. Please change it in a way that makes it more easily playable.

PR #2817, opened 2025-01-24
[Sprig App] Sussy Among Us

doggo742:
### Author name
Author: doggo742

### About your game

**What is your game about?**
Push the boxes to the designated zone- but don't get stuck!!!

wasd, don't push the box into the corner


grymmy:

PR #2820, opened 2025-01-25
[Sprig App] game1

89Aman:
### Author name
Author: 89Aman

### About your game

**What is your game about?**
none

none


Cosmin-Mare (changes requested):
hey! your game appears to be a copy of the tutorial that's not even working. please work on developing your own game and then submit a PR with that

PR #2826, opened 2025-01-27
Kinda fix infinite loop prevention problem

Gamer153:
The current loop prevention can fail for some weird code constructions, like this one:
```js
function __spreadArray(to, from, pack) {
    if (pack || arguments.length === 2) for (var i = 0, l = from.length, ar; i < l; i++) {
        if (ar || !(i in from)) {
            if (!ar) ar = Array.prototype.slice.call(from, 0, i);
            ar[i] = from[i];
        }
    }
    return to.concat(ar || Array.prototype.slice.call(from));
}
```
This happens because the node `path` can have multiple bodies. Typically the main block is in the last entry of the array, which is why I'm choosing the last body node as the real one.

### DIFF
--- a/src/lib/custom-babel-transforms.ts
+++ b/src/lib/custom-babel-transforms.ts
@@ -79,16 +79,18 @@ export default function TransformDetectInfiniteLoop({ types: t }: { types: any }
 				);
 
 				// No block statment e.g. `while (1) 1;`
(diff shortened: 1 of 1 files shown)


grymmy:
Please provide a specific user-facing issue that's caused by this that justifies this change.

grymmy:
Closing this due to lack of commentary/movement.

PR #2832, opened 2025-01-29
[Sprig App] mystery maze 2

scottwellcorn:
### Author name
Author: scottwellcorn

### About your game

**What is your game about?**
this is a game where you need to count your steps to move cause you cannot see your sprite its a combination of mental maths

you can move up, down, left and right there are little differences in the walls


grymmy:
Please reopen this when you have a playable game to present.  In its current form (per Lucas's commentary) it is not playable.

PR #2881, opened 2025-02-12
[Sprig App] Bomber Plane

BoaN235:
### Author name
Author: BoaN235

### About your game

**What is your game about?**
This is a game where the player is a stealth bomber. the players goal is to stop the incoming tanks from attacking the base.

the player has 3 controls move right, move left and fire. you can fire a missile and if a tank is hit it explodes. you gain score evrey time you take down a tank if you fail to take down a tank you will lose health.


grymmy:

PR #2891, opened 2025-02-16
[Sprig App] Tetris Upgraded

Mostlime12195:
### Author name
Author: Mostlime12195

### About your game

**What is your game about?**
Taken inspiration from other Sprig-based Tetris clones, this one is built to have as many features as possible. You can hold S to soft drop, but you can ALSO press L to hard drop your tetrimino! This implementation makes use of *all* 16 available colours, and has beautiful contrasting tetrimonoes, while at the same time, each tetrimino type has the colour that was used when it was first colourised, so 7 uniquely coloured types of blocks! (the cyan tetrimino was replaced with gray because we couldn't use cyan ;( )

Use the A or D keys to move left or right, hold S to soft drop, use W & K to rotate clockwise, use the J key to rotate counter-clockwise, press the L key to hard drop, press I to restart once dead.


grymmy:

PR #2938, opened 2025-03-02
[Sprig App] MAZE RB

Positivecat4:
### Author name
Author: Positivecat4

### About your game

**What is your game about?**
to get to the chest is the main objective and alsoo not to get lost iin the maze

you use w a s d too move the caracter at lvl 1 and al lvl 2 you use i j k l to move the other character the game ends at level 2 when both cracters reach the chest


grymmy (changes requested):
Please add more complexity to this game to justify a Sprig device grant

grymmy:

PR #2952, opened 2025-03-04
Maze game 

Abeer6171:

## Author name

Mohammad Abeer

This is a maze game made using java script

**What is your game about?**

in it we have a player as a red dot and a maze

use arrow keys to move the player

## Code
Java script

## Image (If an image is used)

> Thanks for your PR!


Abeer6171:
I did it wrong so I closed it

PR #2987, opened 2025-03-19
[Sprig App] Graphing Calculator

apinguen:
### Author name
Author: apinguen

### About your game

**What is your game about?**
A calculator that can graph most functions that only use simple operations (addition, subtraction, multiplication, division, exponent)

You can input a number and then select the height/width/scale (spacing of the dots on the axes) by clicking on their respective buttons
Height: Needs to be positive odd integer
Width: Needs to be positive odd integer
Scale: Needs to be positive integer

Non-functions aren't exactly supported; Only one point will be plotted per x value
Only whole numbers supported for points :(
MULTIPLICATION USING TERMS NEXT TO EACH OTHER DOESN'T WORK
(Ex. 2(3); 3x)
You need a multiplication sign in between

L: Switch between graphing/entering equation
WASD: Move cursor
I: Add character
K: Backspace


grymmy:

PR #3008, opened 2025-03-26
[Sprig App Update]

JungleHornet:
This is an update to my game, originally merged in #2883, to fix errors that occur when running on sprig but not in the editor.


grymmy:
lgtm!

PR #3012, opened 2025-03-28
[Sprig App] Just A Game

NicTrixx23-2:
### Author name
Author: NicTrixx23-2

### About your game

**What is your game about?**
Its a very bad game

you use  WASD to move around. at first you can only mine the tree. once you mined it you can press "I" to craft the first tiered pickaxe. with it you can mine Stone. you need to mine all 40 stones to get the last tier of pickaxe, which is crafted by pressing "K". Then you can mine the gemstones and after that you press "L" to finish.


grymmy:
This is a duplicate of #3013

PR #3034, opened 2025-04-09
[Sprig App] MouseRat Game

Lor14N:
### Author name
Author: Lorolol147

### About your game

**What is your game about?**
An escape the maze game where you have to collect the games and escape from the cat

Press "WASD" to move, collect the keys and exit through the square on the left. To restart to the first level press j.


LucasHT22 (changes requested):
Hey! Your game is too similar to other in the gallery. Please make your own game:) If you think this is a mistake, please let us know.

grymmy (changes requested):
Please add more complexity to this game to justify a Sprig device grant

whatwareweb:
I am not sure why, but there are for some reason 2 PRs with the same game, which is making it very confusing.
I'm going to close both PRs, so that you can open a new PR with the correct game that we should review.

PR #3044, opened 2025-04-12
[Sprig App] Pong Ball Simulator

sapbotgit:
### Author name
Author: sapbot

### About your game

**What is your game about?**
It's not much game, it's more lika a pong screensaver.

Sit and watch.


sapbotgit:
I don't think that this demo is worth to be published.

PR #3081, opened 2025-04-24
[Sprig App] Memory Maze

pjontop:
### Author name
Author: pjontop

### About your game

**What is your game about?**
The key objective of the game is to reach the end of the maze. You can see the maze for 3 seconds, and then it will be grayed out. You have to find the end without hitting the wall, or your game is over.

Use the
W - Up
A - Down
S - Right
D - Left


grymmy:
There is a duplicate PR from another person w/ the same game.  Closing both, plz respond here w/ the excuse.

PR #3134, opened 2025-05-20
[Sprig App] Tiger Hun

sapna3863:
### Author name
Author: sapna3863

### About your game

**What is your game about?**
You play as a tiger navigating through a series of complex mazes set in a world ravaged by wildfires. Your goal? To retrieve a mysterious crate and return it safely to a sacred leaf shrine, the last remnant of an unburned world.

Use WASD to move your tiger, and press J to reset each level. Move each box to a leaf, and beware of the fire which will restrict movement.


grymmy:
This seems to be a duplicate.

PR #3190, opened 2025-07-04
[Sprig App] Dungeon Snack

shlok-madhekar:
### Author name
Author: shlok

### About your game

**What is your game about?**
Dungeon Snack is a high-stakes tile-based survival game where you must collect all the coins before time runs out — while being chased by multiple smart slimes. With pushable breakable leaves, limited-time pressure, and strategic movement, this ain’t no casual snack run. Every move matters, and one slip gets you slimed. Think fast, move faster.

Use WASD to move.
Push rocks and leaves to block enemies — but leaves break after one push.
Avoid the slimes chasing you.
Collect all coins before the 30-second timer runs out.
Lose if you run out of time or touch a slime.
Restart happens automatically after a fail.

### DIFF
--- a/games/Dungeon-Snack.js
+++ b/games/Dungeon-Snack.js
@@ -1,0 +1,253 @@
+/*
+@title: Dungeon Snack HARDMODE
+@author: Shlok
+@tags: [maze, ai, collect, chase, hardmode]
+@addedOn: 2025-07-04
+*/
+
+const player = "p"
+const wall = "w"
+const coin = "$"
+const slime = "e"
+const rock = "r"
+const leaf = "l"
+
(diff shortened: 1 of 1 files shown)


GalaxyGamingBoy:
Hi! Your game metadata is missing a `@description` tag. You should add it to let other people in the gallery know what your game is about :) Also the `@tags` array should be with strings, like:
`@tags: ["xyz"]`

whatwareweb:
Hey, i'm going to be closing this as this PR has seen one week without any activity. We'd still love to merge your game, so if you would like us to reopen it, please ping me with @whatwareweb .

PR #3290, opened 2025-08-30
[Sprig App] nagane

thecodedev22:
### Author name
Author: thecodedev22

### About your game

**What is your game about?**
its a simple but fun maze game

You use wasd and collect the gems


whatwareweb:
Hi! Make sure to put the relevant game metadata in your header comment. Here is the format as a reminder :)

Paste this code at the top of your game file fill it out with your game's information where MM is the Current Month (ie 08) and DD is the Current Day (ie 03).
```
/*
@title:
@author: 
@description:
@tags: ["abc", "xyz"]
@addedOn: 2025-MM-DD 
*/
```

Also, it looks like there is some special character in the .js file name, can you change that to only be standard characters?

whatwareweb:
Hey, i'm going to be closing this as this PR has seen one week without any activity. We'd still love to merge your game, so if you would like us to reopen it, please ping me with @whatwareweb .

PR #3324, opened 2025-09-21
[Sprig App] Dungen

Tsunami014:
### Author name
Author: Tsunami014

### About your game

**What is your game about?**
Climb the maze of layers of the procedurally generated dungeon before the time runs out, collecting coins and avoiding traps to escape! But will escape ever come?

WSAD to move, K to reset when you die. The timer ticks down in the top right corner of the screen, when it hits 0 you will be forced to lose and will be shown the map of the current layer to see where the exit was. Get to the stairs to climb to the next level!


whatwareweb:
Hi! Your PR has 3 files, could you please make sure that only one JS game file is added?

Tsunami014:
Is it saying that because I renamed the original file? I thought there ended out being only one

whatwareweb:
Check the files tab here:
https://github.com/hackclub/sprig/pull/3324/files

Tsunami014:
Well I don't know how that managed to happen, I'll just try again

PR #3336, opened 2025-09-29
[Sprig App] Remix of getting_started

whyfren:
### Author name
Author: whyfren

### About your game

**What is your game about?**
this game is about a pixelated player eho is have an objective to set the purple box in a portal, that way we can progres and go fr wining!

W for going up
A for going left
S for going down
D for going right
J for reset the level

basic wasd movement with j for restart, the goal is to set the purple box in the protal.


whatwareweb:
Hi, closing this because this game is too similar to the tutorial. Please make an original game!

PR #3384, opened 2025-10-18
put metadata or missing entries on games that don't have any

JosiasAurel:
self explanatory


whatwareweb:
Thank you for this!!!!!

PR #3422, opened 2025-11-09
Update game metadata and improve metadata checks

whatwareweb:
First commit fixes all the invalid game metadata

Second commit improves error logging of game metadata and adds checks for invalid tags and throws an error with the offending file.


src/integrations/generate-metadata.ts
JosiasAurel, line 1: lgtm

PR #3436, opened 2025-11-13
[Sprig App] Color Switcher

unknowngamer69:
### Author name
Author: unknowngamer69

### About your game

**What is your game about?**
Color Switcher is a small puzzle game where you move around a grid of red and blue tiles. Whenever you step on a tile, it flips color, along with the tiles nearby. Your goal is to change the entire board to one single color. The levels get progressively harder, and every move counts. It’s easy to grasp but genuinely enjoyable to solve.

Use W, A, S, D to move your player.
Each step flips the tile you’re on and the tiles next to it.
Make all tiles become the same color to win the level.
Press R anytime to restart if you get stuck.


unknowngamer69:
Game Name- Color Switcher

PR #4005, opened 2026-06-04
fix website build crashes (script + bad game metadata)

SSoggyTacoMan:
hey! so the website build was failing during the metadata generation step. i noticed a few things:

1. the `mapwithconcurrency` function in the metadata script wasn't handling errors correctly. if even one game file failed to process, the whole build would crash with an unhandled error. i updated it to catch those errors individually and report them properly.
2. `games/Dodge-ball.js` had an invalid metadata block (a typo in the tags and a truncated description). i fixed the typo so it'll pass validation now.
3. `games/Snake-ig.js` had a syntax error in its tags array (`'replayable'dsa`). i fixed the quotes and comma there.
4. `games/Usagi-on-the-dance-floor.js` was missing its author metadata tag. i added it.

should fix the build crashes!


SSoggyTacoMan:
yay i fixed it

PR #4064, opened 2026-07-07
[Sprig App] Hivemind

noahbrown042:
### Author name
Author: Noah B

### About your game

**What is your game about?**
A puzzle game in which you control multiple players at once, featuring 18 levels! (All levels have been thoroughly play tested and are possible.)

Get all the bees onto the flowers to advance to the next level.


LucasHT22 (changes requested):
Nice! What about more mechanics like can't touch in a poison flower? also when i select with J home screen says i tried to select empty map, I know I can use I but would be cool to set map

noahbrown042:
Well I sent that like 3 days before I turned 18, and since I'm not in high school anymore either, I'm no longer eligible to have my game published or to receive a sprig console. If you give me a while, I can still fix that bug and add the hazard mechanic if you'd like.
Edit: Are you sure you aren't using "K" on the "you win" screen? That's the only thing I could find that has a map setting error.

LucasHT22:
Since you started elegible, we probably could send you a sprig!

noahbrown042:
How would I do that?

LucasHT22:
> How would I do that?

do what?

noahbrown042:
I meant how/where would I submit my info to get one?

PR #4068, opened 2026-07-11
[Sprig App] MedDash

mohamadali30907m-source:
### Author name
Author: mohamadali30907m-source

### About your game

**What is your game about?**
medical robot avoid the virus and run to gain the bag

**How do you play your game?**
using the keys: A,D  W,S

PR #4069, opened 2026-07-11
[Sprig App] MedDash

mohamadali30907m-source:
### Author name
Author: Mohamad Ali Ahmed

### About your game

**What is your game about?**
MedDash is an arcade-style survival game where the player controls a medical robot inside a high-tech space laboratory. The goal is to collect vital supply bags while constantly evading a self-mutating smart virus that actively tracks the player's movements.

- Use the 'W' key to move Up.
- Use the 'S' key to move Down.
- Use the 'A' key to move Left.
- Use the 'D' key to move Right.
- Collect the hospital bags to increase your score (Score +10 per bag).
- Avoid touching the virus! You only have 3 lives (HP).
- Reach 100 points before the 60-second timer runs out to secure VICTORY!

### DIFF
--- a/games/MedDash.js
+++ b/games/MedDash.js
@@ -1,0 +1,182 @@
+/*
+First time? Check out the tutorial game:
+https://sprig.hackclub.com/gallery/getting_started
+
+@title: MedDash
+@description: medical robot avoid the virus and run to gain the bag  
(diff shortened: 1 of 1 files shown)


LucasHT22:
Check #4068  for your review!

PR #4129, opened 2026-08-13
fix(sequencer): prevent NaN BPM from freezing the tune editor (#2823)

### LINKED ISSUE #2823: Tune editor glitches when changing BPM on a tune with closed grave accents
If you use the built in Tune editor to edit a Tune which has a closed grave accent pair (like so:) Image the Tune editor stops working as intended. In paticular, it causes many of its visual aspects to stop working, such as it not displaying any of the insturments added after editing it, the play bar not moving when playing the track, and the BPM number not updating with the slider. Most concerningly though, it prevents you from being able to exit out of the editor via the close button, which from what I can tell effectively forces you to refresh the page to escape.

If one wants to reproduce this bug, just create a new tune with a pair of close grave accents, open the editor and change the bpm slider, then try to do something else and you should immediately see issues. 

Evidence (appologizes if its a bit frantic, if requested I can maker another):
https://github.com/user-attachments/assets/0cf0c467-3712-48be-a88c-3e9647505adc

I am unaware of any work arounds other then not closing the grave accent when editing it, but that itself causes the tune function to overwrite code beneath it with itself.

The only other thing I can think of adding is that this might have a connection with the bug listed in https://github.com/hackclub/sprig/issues/430 ? It might be worth you fellas checking it.

Best of luck fixing this issue, and please let me know if you need any additional info or clarification.

kawacukennedy:
Fixes #2823

## Problem

When a tune contains closed grave accents (backticks), e.g. from pressing backtick-close in the editor, the tune text is `` `400: C4~400,\n11600` ``. The BPM is parsed from the text via `/(.+):/` + `Number(...)`, which yields `NaN` for `` `400 ``. That NaN is stored in `bpm.value` and passed into `play()` where `setTimeout(go, NaN)` fires immediately, creating an infinite loop that freezes the tab (cannot exit the editor, requires a page refresh).

## Fix

Editor-side only; the game runtime (`textToTune`/`tuneToText`) is untouched.

- `sequencer-utils.ts`: add a `getBPM(text)` helper that falls back to a finite default (120 BPM) whenever the parsed BPM is non-finite or non-positive.
- `sequencer-utils.ts`: guard `playBeat` and `play()` against non-finite BPM / note durations, and fall back to a sane timeout delay.
- `sequencer.tsx`: use `getBPM(...)` when parsing the tune text into cells.

Once `bpm.value` is finite, the editor's own sanitizing effect strips the backticks, so the tune text round-trips cleanly (`500: C4~500,\n15500`) instead of corrupting to `NaN: C4-NaN,\nNaN`.

## Tests

Adds `sequencer-utils.test.ts` with 6 unit tests covering the NaN BPM case, malformed input, and clean round-trips. Verified locally with the repo's pinned toolchain:

- 6/6 pass with the fix
- 5/6 fail against the pre-fix code (confirming they catch the bug)
- `tsc --strict` clean


kawacukennedy:
This is a bug fix for #2823 (editor-side, not a game submission), so the `submission` label/validation doesn't apply here.

The CI runs (Unit Tests, Astro Build, Docker Build, Build Spade UF2) are waiting on workflow approval for this fork PR, and the Vercel deploy needs a Hack Club Team member to authorize it. Could a maintainer approve those? Thanks!

kawacukennedy:
Update: rebased onto current `main` (`c2b8081`) — no conflicts — and force-pushed as `eff2481a`.

About the earlier CI notices: those runs show "No jobs were run" because they sat pending workflow-approval from a maintainer (required for fork PRs) until they expired — they never executed, so they're not test/build failures. Fresh runs for `eff2481a` are queued and again awaiting approval.

FWIW, I verified the unit suite locally with the repo's pinned toolchain (vitest 4.1.7 / vite 7.3.2, `@babel/standalone` 7.29.x, and the CI env vars `PUBLIC_MAX_ITERATIONS=2000` `PUBLIC_MAX_LOOP_TIME_MS=1500`): `sequencer-utils.test.ts` 6/6 and `error.test.ts` 11/11 pass.

Could a maintainer approve the pending workflow runs (and the Vercel deploy)? Thanks.

PR #4169, opened 2026-09-02
Point retired CDN asset URLs at /rescue (49 links)

jeninh:
Points **49 retired-CDN asset references** across **17 files** at the current CDN's lookup endpoint, `cdn.hackclub.com/rescue?url=<original>`.

The old Vercel and Hetzner origins are being retired. `/rescue` resolves an original URL to wherever that file now lives on the v4 CDN, so wrapping the existing URL keeps these assets loading without moving any bytes or re-uploading anything.

```diff
- ![screenshot](https://cloud-abc123-hack-club-bot.vercel.app/0image.png)
+ ![screenshot](https://cdn.hackclub.com/rescue?url=https://cloud-abc123-hack-club-bot.vercel.app/0image.png)
```

### Scope

- 49 references rewritten across 17 files.
- 16 references were **already** wrapped in `/rescue` and are left exactly as they were.
- Text references only — no image bytes, no filenames, no uploads, no layout changes.
- Every change is a pure URL-for-URL substitution, verified programmatically: masking the rewritten spans in both the original and the result yields byte-identical documents, so no surrounding markup was touched.
- Original URLs are preserved verbatim inside the `url=` parameter, so this is fully reversible and nothing is lost.


LucasHT22 (approved):
Awesome! TYSM!

PR #4181, opened 2026-09-06
[Sprig App] Last Battery

Qazi-01:
### Author name
Author: Tashfeen Miyar

### About your game

**What is your game about?**
When I heard about sprig I just knew I wanted it so I made last battery a maze survival game where you have to sneakily reach the door avoiding the drones that follow you. Collect batteries to charge and move forward. The game is simple for now visually and structurally. The drone mechanics were hard to get right printing text remained a problem through out as I'm a symmetry kind of guy. I don't know if there's a game like this already available on sprig console I just really enjoyed playing the game myself that I think its unique the originality is something.

Use WASD to move and I to restart when you lose. The game is simple just try to reach the doors, collecting batteries and avoiding the drone. The game is endless the score increases on each escape.


Qazi-01:
Metadata difference fixed.

Qazi-01:
The exit bitmap worked before but I made it 16x16 just to ne safe and works even now. The for loop spacing issue was marked addressed as I didn't found it in my original game on sprig website also it might be in the new suggestion commits from copilot. I also saw above that submission label not added but I am unable to add it or I just don't know how.

PR #4200, opened 2026-09-09
[Sprig App] Pirate bay

gavelleleberrec-tech:
### Author name
**Author:** gavelleleberrec-tech

### About your game
**What is your game about?**
A slime escaping sinking pirate ships. Use powerups to aid you in your journey and your sword to cut tentacles

WASD, and I to attack


LucasHT22 (changes requested):
Nice! What about a restart key?

PR #4235, opened 2026-09-20
[Sprig App] my_roborock_game

ej2508:
### Author name
Author: ej2508

### About your game

**What is your game about?**
Robo Rock is a fast, simple arcade game where your goal is to keep your robo alive and score as many points as you can. Rocks fall from the top of the screen in random lanes, one row at a time, and you steer the robo with W, A, S and D to slip past them. You can move in all four directions but can't leave the play area, so you can dodge by dropping back as well as by sliding sideways. Every rock that falls past you without touching your robo earns 1 point. The game ends the moment a rock lands on your robo, or you move into one, and your final score is shown on screen. Because the rocks appear in random lanes, no two rounds play the same, so you have to react instead of memorizing a pattern. The robo and rock sprites are drawn by me in the Sprig pixel editor.

Change the last sentence if it isn't true, for example if you've only drawn one of the two sprites. If you add a twist later, like rocks that speed up as your score grows, add a sentence about it before the last one. The description should only promise what the game does.

Steer the robo with W, A, S and D. It can move up, down, left and right. so simple isn't?


SSoggyTacoMan:
hi! looks like this is a duplicate of #4236, please close this one!

ej2508 (commented):
metadata added

ej2508:
> hi! looks like this is a duplicate of #4236, please close this one!

Sorry, i am first to this. thank you for point out

PR #4252, opened 2026-09-21
fix: accept flexible author formats in PR description

SSoggyTacoMan:
Submitters often write their name without bold syntax, wrap it in a markdown link, or write @username. Currently extractBoldField strictly expects **Author:**, which causes false validation failures even when the field is filled.

This update allows:
- Unbolded Author: name
- Fully bolded **Author: name**
- Markdown links [@user](url) and mentions @user
- List items like - Author: name


LucasHT22 (approved):
LGTM

